PR #1, opened 2022-03-10
Added reference case studies from OS paper

sgiarols:
# Description

Inclusion of additional case studies for user reference

Fixes # (issue)
#2 

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Why are there some tests that fail and then, when they are re-run, they pass? It doesn't look very estable.

PR #22, opened 2022-03-24
Version

### LINKED ISSUE #21: Model version to 1.0.0
Update the model version from 0.9.2 to 1.0.1

sgiarols:
# Description

This PR upgrade to 1.0.0 version, after reconciling version progression.

Fixes #21 

## Type of change

## Key checklist

## Further checks


setup.cfg
```diff
@@ -1,5 +1,6 @@
 [bumpversion]
-current_version = "1.0"
+current_version = 1.0.0
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<release>.*)-(?P<build>\d+))?
```
dalonsoa, line 3: What is this for?

src/muse/demand_matching.py
```diff
@@ -264,7 +264,7 @@ def _demand_matching_impl(
     dimensions have coordinates. Input sanitization is performed in `demand_matching`
     proper.
     """
-    from numpy import isnan, prod
+    from numpy import all, isnan, prod
```
dalonsoa, line 267: Why are there changes in this PR about demand matching and timeslices when it is supposed to be only about the version?

dalonsoa (changes requested):
I think this PR mixes unrelated things. If it is about updating the version, that's the only thing it should contain. Otherwise it is confusing.

setup.cfg
dalonsoa: As far as I can tell, this is only useful if you are planing to have an advanced release versioning scheme, in which case it is useful to define [bumpversion:part:release] and [bumpversion:part:build] sections in the `setup.cfg`. I really suggest you keep this simple until you get familar with versioning.

sgiarols: Parse is needed to get the version changed and the version parsed correctly

Issue #33, opened 2022-04-19
Cache quantities: tutorial


sgiarols:
Add a tutorial to the documentation about how to define a new property to cache.

PR #35, opened 2022-05-12
Remove retro

dalonsoa:
# Description

Adds a new version for the demand share that do not use retro agents - indeed, if they are present in the `Agents.csv` file, it will complain. Using this new demand share is set in the TOML file for each of the sectors. 

Fixes # NA

## Type of change

## Key checklist

- [x] The documentation builds and looks OK: `$ python setup.py build_sphinx`

## Further checks

### DIFF
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -143,6 +143,10 @@ def create_newcapa_agent(
         for name in kwargs.pop("search_rules")
     ]
 
+    if not retrofit_present:
+        if "with_asset_technology" not in search_rules:
+            search_rules.insert(-1, "with_asset_technology")
+
     result = InvestingAgent(
         assets=assets,
         region=region,
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -91,6 +91,7 @@ def create_newcapa_agent(
     capacity: xr.DataArray,
     year: int,
     region: Text,
+    share: Text,
     search_rules: Union[Text, Sequence[Text]] = "all",
     interpolation: Text = "linear",
     merge_transform: Union[Text, Mapping, Callable] = "new",
@@ -120,7 +121,6 @@ def create_newcapa_agent(
             capacity.sel(asset=existing.values, year=years)
         )
     else:
-        share = "agent_share_2"
         technologies = kwargs["technologies"]
         assets["capacity"] = _shared_capacity(
             technologies, capacity, region, share, year, interpolation=interpolation
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -125,6 +125,7 @@ def create_newcapa_agent(
         assets["capacity"] = _shared_capacity(
             technologies, capacity, region, share, year, interpolation=interpolation
         )
+        merge_transform = "merge"
 
     kwargs = _standardize_investing_inputs(
         search_rules=search_rules,
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -96,9 +96,14 @@ def create_newcapa_agent(
     merge_transform: Union[Text, Mapping, Callable] = "new",
     quantity: float = 0.3,
     housekeeping: Union[Text, Mapping, Callable] = "noop",
+    retrofit_present: bool = True,
     **kwargs,
 ):
-    """Creates newcapa agent from muse primitives."""
+    """Creates newcapa agent from muse primitives.
+
+    If there are no retrofit agents present in the sector, then the newcapa agent need
+    to be initialised with the initial capacity of the sector.
+    """
     from muse.filters import factory as filter_factory
     from muse.registration import name_variations
 
@@ -108,8 +113,18 @@ def create_newcapa_agent(
     existing = capacity.interp(year=year, method=interpolation) > 0
     assert set(existing.dims) == {"asset"}
     years = [capacity.year.min().values, capacity.year.max().values]
+
     assets = xr.Dataset()
-    assets["capacity"] = xr.zeros_like(capacity.sel(asset=existing.values, year=years))
+    if retrofit_present:
+        assets["capacity"] = xr.zeros_like(
+            capacity.sel(asset=existing.values, year=years)
+        )
+    else:
+        share = "agent_share_2"
+        technologies = kwargs["technologies"]
+        assets["capacity"] = _shared_capacity(
+            technologies, capacity, region, share, year, interpolation=interpolation
+        )
 
     kwargs = _standardize_investing_inputs(
         search_rules=search_rules,
@@ -272,6 +287,11 @@ def agents_factory(
         if capacity.dst_region.size == 1:
             capacity = capacity.squeeze("dst_region", drop=True)
     result = []
+
+    retrofit_present = False
+    for param in params:
+        retrofit_present = retrofit_present or param["agent_type"] == "retrofit"
+
     for param in params:
         if regions is not None and param["region"] not in regions:
             continue
@@ -283,7 +303,7 @@ def agents_factory(
         param["capacity"] = deepcopy(capacity.sel(region=param["region"]))
         param["year"] = year
         param.update(kwargs)
-        result.append(create_agent(**param))
+        result.append(create_agent(**param, retrofit_present=retrofit_present))
 
     nregs = len({u.region for u in result})
     types = [u.name for u in result]
--- a/src/muse/demand_share.py
+++ b/src/muse/demand_share.py
@@ -63,6 +63,17 @@ from mypy_extensions import KwArg
 from muse.agents import AbstractAgent
 from muse.registration import registrator
 
+
+class InconsistencyInDemandShare(Exception):
+
+    msg = """A retrofit agent has been found in a 'New agents'-only demand share
+function. Make sure you remove all the retro agents from the Agents input files or use a
+demand share method that can handle both new and retro agents."""
+
+    def __str__(self):
+        return self.msg
+
+
 DEMAND_SHARE_SIGNATURE = Callable[
     [Sequence[AbstractAgent], xr.Dataset, xr.Dataset, KwArg(Any)], xr.DataArray
 ]
@@ -82,10 +93,10 @@ def factory(
     settings: Optional[Union[Text, Mapping[Text, Any]]] = None
 ) -> DEMAND_SHARE_SIGNATURE:
     if settings is None or isinstance(settings, Text):
-        name = settings or "new_and_retro"
+        name = settings or "split_demand"
         params: Mapping[Text, Any] = {}
     else:
-        name = settings.get("name", "new_and_retro")
+        name = settings.get("name", "split_demand")
         params = {k: v for k, v in settings.items() if k != "name"}
 
     function = DEMAND_SHARE[name]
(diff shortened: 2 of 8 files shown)


src/muse/demand_share.py
```diff
@@ -314,6 +325,89 @@ def decommissioning(capacity):
...
+    technologies: xr.Dataset,
+    production: Union[Text, Mapping, Callable] = "maximum_production",
+    current_year: Optional[int] = None,
+    forecast: int = 5,
+) -> xr.DataArray:
+    r"""Splits demand across new agents.
+
+    The input demand is split amongst *new* agents. *New* agents get a
+    share of the increase in demand for the forecast yearas well as the demand that
+    occurs from decommissioned assets.
```
sgiarols, line 341: years (instead of yearas)

src/muse/demand_share.py
```diff
@@ -63,6 +63,17 @@ def demand_share(
...
+class InconsistencyInDemandShare(Exception):
+
+    msg = """A retrofit agent has been found in a 'New agents'-only demand share
+function. Make sure you remove all the retro agents from the Agents input files or use a
+demand share method that can handle both new and retro agents."""
+
+    def __str__(self):
+        return self.msg
+
+
```
sgiarols, line 76: If we only remove the "retrofit" agents from Agents.csv, we get this error:

 File "c:\users\sg2410\desktop\github\muse_os\src\muse\demand_share.py", line 463, in _inner_split
    total = sum(shares.values()).sum("asset")  # type: ignore
AttributeError: 'int' object has no attribute 'sum'

We may add this to the error lists with a message like:

"This error refers to an agent with no assets. To fix this error, check the capacity assigment to the agents. ONe possibility is that you have decided not to use "Retrofit" agents, as such you may have already removed them from the agent definition file and the file of technodata, the system TOML file should change the demand_share to "standard_demand" function in each subsector section for each of the selected sectors to model.

sgiarols (changes requested):
- The assets are not initialised for the New agents. In the base year, the new agents should have the same assets as the retrofit.

- It may be useful to add a test on the __init__() function

https://github.com/SGIModel/MUSE_OS/blob/69a2bac60d6d19388094886ea47d993e975f0413/src/muse/agents/agent.py#L14

 to check that assets and all the attributes are properly initialised as for the retrofit agent (apart from the demand_share function) in the first year of the simulation.

- A new type of ineractions "new_to_new" should need to be recorded. This would take out the printout "subsector_name.interaction_name" and may probably guarantee that the assets should be inhertited and trasmitted properly form one period to the next after the investment is made.

src/muse/demand_share.py
sgiarols: The steps to remove retro ahould be documented

dalonsoa:
Just to make sure I understand this right: when there is only `newcapa` agents in a given sector, the following line when creating the newcapa agent:

https://github.com/SGIModel/MUSE_OS/blob/69a2bac60d6d19388094886ea47d993e975f0413/src/muse/agents/factories.py#L112

Should be replaced by something similar to this one, taken from the retrofit creation, so the newcapa agents actually owns the assets:

https://github.com/SGIModel/MUSE_OS/blob/69a2bac60d6d19388094886ea47d993e975f0413/src/muse/agents/factories.py#L70

Is this correct?

sgiarols:
yes the interpretation makes sense.

dalonsoa:
So, how do you see this moving forward? Assuming all the code is ok - still to be seen after your review - For even the simplest model to run we will need to increase the cap in the mac capacity expansion in the technologies files, will surely affect other models.

dalonsoa:
I'm not sure what do you mean. The calculation does not pass the first year because of the constrains, as I described above, so I don' know how you got until 2050. Could you elaborate?

sgiarols:
I have increased the growth constraints in these runs, so the simulation goes to the end

sgiarols:
[default.zip](https://github.com/SGIModel/MUSE_OS/files/8785573/default.zip)
This is the case study I have been running.

sgiarols:
[Technodata.zip](https://github.com/SGIModel/MUSE_OS/files/8785797/Technodata.zip)

dalonsoa:
This model is not running for me using the same version of the PR. It complains of missing Retro agent. Did you change anything else in order to run it?

sgiarols:
nothing I can think of. can you paste the full error?

dalonsoa:
```
Exception has occurred: RetrofitAgentNotDefined
The retrofit agent has not been defined. This might be because it actually
has not been defined in the agents file or specified in the technodata file or it cannot
be found because there is a typo in its name in either the agents and or technodata
files. When a retrofit agent is defined in the agent file, the same name should be
reported as an additional column in the technodata, assigning the fraction of the
initial stock assigned to each retrofit agent. In presence of multiple retrofit agents,
a new column needs to be added per agent.

During handling of the above exception, another exception occurred:

  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/agents/factories.py", line 340, in _shared_capacity
    shares = technologies[share].sel(technology=capacity.technology)

During handling of the above exception, another exception occurred:

  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/agents/factories.py", line 342, in _shared_capacity
    raise RetrofitAgentNotDefined
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/agents/factories.py", line 125, in create_newcapa_agent
    assets["capacity"] = _shared_capacity(
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/agents/factories.py", line 170, in create_agent
    return method(**kwargs)  # type: ignore
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/agents/factories.py", line 310, in agents_factory
    result.append(create_agent(**param, retrofit_present=retrofit_present))
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/sectors/subsector.py", line 184, in factory
    agents = agents_factory(
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/sectors/sector.py", line 54, in <listcomp>
    Subsector.factory(
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/sectors/sector.py", line 53, in factory
    subsectors = [
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/mca.py", line 74, in factory
    sectors.append(SECTORS_REGISTERED[kind](sector, settings))
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/__main__.py", line 52, in muse_main
    MCA.factory(settings).run()
  File "/Users/dalonsoa/Documents/Projects/MUSE_OS/src/muse/__main__.py", line 55, in <module>
    muse_main()
```

sgiarols:
I see why this happened - because I had not yet incorporated the last changes. something in the last commits seem to get not recognised the absence of retrofit agents

src/muse/agents/factories.py
```diff
@@ -96,9 +96,14 @@ def create_newcapa_agent(
     merge_transform: Union[Text, Mapping, Callable] = "new",
     quantity: float = 0.3,
     housekeeping: Union[Text, Mapping, Callable] = "noop",
+    retrofit_present: bool = True,
     **kwargs,
```
sgiarols, line 100: is it this one put to False afterwards?

src/muse/agents/factories.py
```diff
@@ -108,8 +113,18 @@ def create_newcapa_agent(
...
+    if retrofit_present:
+        assets["capacity"] = xr.zeros_like(
+            capacity.sel(asset=existing.values, year=years)
+        )
+    else:
+        share = "agent_share_2"
+        technologies = kwargs["technologies"]
+        assets["capacity"] = _shared_capacity(
+            technologies, capacity, region, share, year, interpolation=interpolation
+        )
```
sgiarols, line 128: This line forces to bring back the agent_share_2 line. It may need to be changed into smething more flexible, what if we have more than 1 agent?

sgiarols: To run the model, I had to change on line 123, "2" to "1". 
However, we may follow the same approach used for the standard_agent, here

`https://github.com/SGIModel/MUSE_OS/blob/9bf1957c28b0030a442f3bc24353f10943bc64d6/src/muse/agents/factories.py#L12`

sgiarols:
@dalonsoa 
now with the different initialisation, I do not have the capacity bounding, but the capacity does not meet the demand (without the algorithm failing) and the initialisation of the agent is too case specific as it requires agent_share_2 to be defined.

dalonsoa:
Ok, one thing at a time. You are actually right, but just let's make sure first we are on the same page.

I'm using as starting point to create a model that uses no retro agents the "default" example (this one: https://github.com/SGIModel/MUSE_OS/tree/develop/run/example/default). I changed two things there:
- Remove the retro agent row from "`Agents.csv`.
- Use `standard_demand` in the relevant places of the `settings.toml` file. 

Now, if I compare this with the model you have sent, I've found you also did this but, in addition there is also a difference in the `Technodata.csv` files. There you have only "Agent1", assigned to "New" for Power and Gas, but not for Residential where you had only "Agent2" assigned to "Retrofit". I think you probably just sent me the wrong Technodata file for residential.

So, if I fix the tehcnodata file file for Residential so it only has Agent1 with the thechnologies assigned to it and - as you point out - change `agent_share_2` by `agent_share_1`, then things run. 

Now, let's see if I can get rid of the hardcoding of `agent_share_1`.

dalonsoa:
@sgiarols , I think I've solved the problem. `share` was not enabled for `newcapa` agents - so even if you had several, they will not share the demand. I've enabled that by commenting the relevant liens in the readers and removed the hardcode of the agent share. 

Could you try now and see if (1) it runs and (2) the results make sense?

sgiarols:
I am running the defalt case and I get the bumps again
image

dalonsoa:
Ok, I think I've sorted out the issue. The problem was that `newcapa` agents have as default method for merging existing and new assets `new`, meaning that existing assets are discarded. That is not a problem normally because `newcapa` agents have no existing assets - they have been transferred to the `retrofit` agents via the interaction networks. However, in this case, it is a problem: we do want to keep existing assets. The solution was as simple as setting`merge_transform = "merge"` when creating the `newcapa` agents in the case when there is no `retro (in [this line](https://github.com/SGIModel/MUSE_OS/blob/bae1f7a1aa9415edb2d932118bc07ed2d474a4b6/src/muse/agents/factories.py#L128)). Fetch the last version of this PR to get the solution.

When that is done, the model runs all the way to the end, populating `MCACapacity.csv` and the other files with a sensible number of entries. Numerically it is a bit different from the case of having `retro` agents, but I think that is to be expected given that the constraints for the solver are different as long as the trends are the appropriate ones.  

I attach the `default` model - renamed `default_original` taken from `src/muse/data/example` - and exactly the same model without retrofit agents, renamed `default_no_retrofit`. I only had to increase the `MaxCapacityAddition` and `MaxCapacityGrowth` in the Residential technodata, and only by a factor of 2 and 2.5 respectively. For Power and Gas it was OK. The corresponding Results folders contain the actual results, so you can compare the trends without the need of running them.

[comparison.zip](https://github.com/SGIModel/MUSE_OS/files/8806617/comparison.zip)

sgiarols:
The changes make sense to me. 
Most of the time removing the retro does not change the results if it is a one-sector model or a multiple sector agent with the same objective. The attached mode, a three sector case with two agents, having two different objective functions, show two different results. Some situations may trigger just enough of a price change to induce a different technology section. In fact, the agents are investing using different capacities, which would affect the decisions.  So it seems reasonable that removing the retrofit agent results in a different model.
[modified.zip](https://github.com/SGIModel/MUSE_OS/files/8893197/modified.zip)

We need to check the failing tests during our call.

sgiarols:
Some fail ebcause the technodata in the tests does not contain "share". 
I interpret that the non investng aggregating agents do not use the share
image

sgiarols:
Another one looks for agent_share 1
image

sgiarols:
Part 1) I ran the trade case study.  The trade case runs with the folder here attached (only the residential needs to have the standard demand; power and gas are "dafault" agents which do not use retrofit). The trade default agents use a demand_share = "unmet_forecasted_demand", which does not account for the retrofit, they still need the agent share 
(you can see it from 
https://github.com/SGIModel/MUSE_OS/blob/develop/src/muse/data/example/trade/technodata/power/Technodata.csv
)
to distribute the capacity among the agents. However, in a pre-step the trade runs "nonInvestingAgent" which do not invest but updates metrics and do not need the share. THis is why the error is triggered

Part 2) Capacities.csv has, as I see, an error in the output. Apart from this reporting error, checking the outputs in the folders, they are pretty similar but in 2030 there is a 5 PJ difference in the residential sector. Here I can see an action about checking the capacity gap over time. I can try and do this.

Part 3) If we use standard_demand, the model would not complain about having the interaction section new_to_retro in the folder attached, I commented it, but maybe we should trigger an error?

Part 4) It seems we can rename the section in the toml, for example  I changed "new_to_retro" into "standard" in the toml section   [sectors.residential.subsectors.standard], and there is no complain about it. So I assume there is no consequence on the simulation.
[trade.zip](https://github.com/SGIModel/MUSE_OS/files/8902225/trade.zip)

dalonsoa:
Great! Thanks to tracking this down. I'll check 1) asap - possibly tomorrow. About 3 and 4, if there are no consequences, I'll leave it as it is for now and just change it when updating the documentation, as it is most likely a question of avoiding confusion for the users. I look forward to see what happens with 2. Please, let me know of the outcome.

sgiarols:
@dalonsoa 
I found that the default agent does not interpolate the "ExistingCapacity" appropriately. I therefore opened an issue about it. This is causing a dscrepancy for the residential sector. If I complete the decommissioning profile for each modelled year, then the residential sector converges to the same results. I opened an issue on this here
https://github.com/SGIModel/MUSE_OS/issues/46

I think that the current PR can be closed provided that the test causing the error for the 'non investing agent" is solved.

Then the documentation can be completed.

dalonsoa:
@sgiarols , this is finally passing. Please, review this and I'll start to work on the documentation (new issue open #47 ), as well as in the #33 .

sgiarols (approved):

PR #118, opened 2023-05-19
Convert setup.cfg to pyproject.toml and add entrypoint.

### LINKED ISSUE #103: Move information from the legacy setup.cfg to pyproject.toml
While definition of the python package and other functionality using `setup.cfg` is perfectly fine, is becoming the legacy approach in favour of `pyproject.toml`. 

While MUSE has some information in `pyproject.toml`, it is mostly empty and all the information is still in `setup.cfg`. This should change. A guide of what a `pyproject.toml` file should look like with all the relevant sections can be found in the [setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). 

Note that there are other tools using `setup.cfg` that should be moved, and others that cannot be moved because they don't know how to use `pyproject.toml` yet (e.g. flake8, which should be moved to its own `.flake8` file).

### LINKED ISSUE #108: Convert muse into a command
After #103 

Convert muse into a command, so you can use `muse settings.toml` for running MUSE instead of `python -m muse settings.toml`.

To do this, [an appropriate entry point in `pyproject.toml`](https://setuptools.pypa.io/en/latest/userguide/entry_point.html) must be created pointing (probably) to `muse/__main__.py`, but this will need to be confirmed. 

While it seems like a minor change, the above will enable the use of [pipx](https://pypa.github.io/pipx/), keeping MUSE within its own virtual environment (without the user needing to create it manually), simplifying future updates and allowing it to be used anywhere in the OS.

jamesturner246:
# Description

Moved most of that is possible from setup.cfg into pyproject.toml.

Added an entry point via src/muse/main.py to pyproject.toml. MUSE can now be called via `muse` command after installation

bumpversion and flake8 do not yet support pyproject.toml, and remain in setup.cfg.

NEEDS INFO: `bump2version` fails doe to incorrect initial version in setup.cfg bumpversion

Fixes #103 
Fixes #108 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,4 +2,127 @@
 requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "MUSE"
+authors = [
+    {name = "Sustainable Gas Institute", email = "sgi@imperial.ac.uk"},
+]
+description = "Energy System Model"
+readme = "README.md"
+requires-python = ">= 3.7"
+keywords = ["energy", "modelling"]
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Programming Language :: Python :: 3.8",
+    "Intended Audience :: Science",
+    "Intended Audience :: Research",
+    "Intended Audience :: Economists",
+    "Intended Audience :: Energy Experts",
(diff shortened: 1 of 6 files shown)


jamesturner246:
Fixes #108

jamesturner246:
Got it. I have pre-commit installed now. The more you know..

PR #119, opened 2023-05-19
Expand CI workflow

### LINKED ISSUE #104: Add testing in the CI system to cover Python 3.8-3.11, Linux, Windows and MacOS
Currently, MUSE only supports, officially, Python 3.8. But that's several years old and that limitation is very restrictive. Moreover, it only run tests in Windows. This issue aims to:

- Setup the CI system (GitHub Actions) to run the tests in Python 3.8-3.11
- Setup the CI system to run tests in all three platforms: Linux, Windows and MacOS

This can be done all at once with a [job matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs). 

As, most likely, some combinations of the above will not work altogether (eg. some dependency does not work in a Python version/platform combination) or will make some tests fail:

- Open an issue for each of the failing combination, indicating the tests that fail and including the log output.
- [Exclude that combination in the testing matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations)

TinyMarsh:
# Description

This PR expands the CI system to run the tests in Python 3.8-3.9 as well as in all three platforms: Linux, Windows and MacOS. 

Close #104 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
     pull_request:
         types: [opened, synchronize, reopened]
     push:
-        branches: [main, develop]
+        branches: [main, develop, expand_ci_system]
 
 jobs:
   # Checks the style using the pre-commit hooks
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,8 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        fail-fast: false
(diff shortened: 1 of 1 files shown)


dalonsoa:
@TinyMarsh , unit tests with Python 3.9 in Windows is failing. I'm re-running the CI, in case it was a flaky test, but if it indeed fails, then we need to open an issue about it and remove Python 3.9 from the list (or indicate somewhere that it works in Python 3.9 but only in MacOS and Linux).

PR #121, opened 2023-05-19
Updating pyproject.toml with valid python versions

TinyMarsh:
# Description

This PR addresses #105 updating the `pyproject.toml` file to include the Python versions which were included in the CI workflow in #119 . Specifically, the valid versions that MUSE can now be installed in are 3.8 and 3.9.

Closes #105 

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking, back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (whatever its nature)

## Key checklist

- [x] All tests pass: `$ python -m pytest`
- [ ] The documentation builds and looks OK: `$ python setup.py build_sphinx`

## Further checks

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works

Issue #120, opened 2023-05-19
Revise the version of Pandas being targeted


TinyMarsh:
While experimenting with extending the CI workflow (#119) to test on a matrix of python versions (3.8 - 3.11) and platforms (Linux, Windows and MacOS), tests failed for Python versions 3.10 and 3.11 on all platforms.

The cause of failure in each of these combinations was the targeting of `pandas<=1.3`. As per [the documentation](https://pandas.pydata.org/pandas-docs/version/1.3/getting_started/install.html#python-version-support), Pandas 1.3 does not support Python versions 3.10+. The minimum version of Pandas which supports up to Python 3.11 is [Pandas 1.5](https://pandas.pydata.org/pandas-docs/version/1.5/getting_started/install.html#python-version-support).

Suggest investigating if MUSE's dependency of `pandas<=1.3` is still relevant and, is possible, switch to targeting `pandas>=1.5`.

sgiarols:
Thanks. I often regret that we moved to xarray, rather than just sticking to pandas, as xarray constantly undergoes massive updates. I think we need the tool should be sustained in the long-term and that an upgrade to the newest packages seems crucial at this stage. I would leave to @ahawkes final considerations, as the refactoring time, if we decide to go for it, may require a conversation on the engagement plans.

dalonsoa:
I think `xarray` is the right tool for the job, but it is true that we picked it when it was in a very early state of development and it was not stable. That was possibly an error, indeed. The update in 2022.6.0 has annoyed a lot of people, but I think it has made the tool more mature and things should be more mature from now on.

Thinking in the future, I think it is important not to kick the can down the road and when incompatibilities arise with any new version of any package, we fix them rather than pinning to an older version because that only makes the tool to accumulate technical debt until it explodes.

Anyway, let's thing about it and maybe we can have a chat about how to best proceed.

dalonsoa:
I've already started to work on this. Fixed a couple of issues, but I'm stuck in another. I've posted a question in StackOverflow in case someone can point me in the right direction: https://stackoverflow.com/questions/76471238/why-i-cannot-add-a-dataarray-to-an-existing-dataset-with-a-multiindex

dalonsoa:
[...]

Issue #109, opened 2023-05-19
Update installation instructions for windows


dalonsoa:
As of today, as far as I know, none of the current MUSE dependencies need to be built from source, so [the note for Windows users in the instructions could be removed](https://muse-os.readthedocs.io/en/latest/installing-muse.html) - or moved to a less prominent troubleshooting section to avoid confusion.

dalonsoa:
I'm not totally sure I understand. Is it that the point about installing build tools in Windows is correct for 3.11 but no longer needed for 3.8? 

I'm not entirely surprised. If I remember correctly, those tools were needed because there were no binary wheels for some dependencies under Windows, but I guess that is no longer the case. For what we can see in Ryan's PR, all tests run with Python 3.9, so I guess it is not a problem there either. On Python 3.10, it is to be seen, I guess.

dc2917:
Yes, exactly. 

So I'm wondering whether it's best to either:

1. remove the section about installing Microsoft Visual C++ etc from the documentation and state the requirement to use a Python version where they're not needed, or 
2. let users use whichever Python version they want, but state that they might need Microsoft Visual C++ to build some dependencies if they're on a recent Python version.

dalonsoa:
Let's see what happens with #120. If removing the pinning of pandas (and sorting any other incompatibility) means that build tools are not needed, then I'd go for it.

dalonsoa:
Going back to this one, I think option 2 is better. I'll update the documentation to reflect that an appropriate Python version is required (and providing guidance on how to do so), and remove any reference to MSVC++. It's one thing less to install.

Issue #106, opened 2023-05-19
Deploy MUSE versions to PyPI


dalonsoa:
After #105 

Deploying MUSE to PyPI will massively simplify its distribution and installation by end users, as well as enable other potential uses down the line. 

[Here there is a nice guide](https://dev.to/iamtekson/publish-package-to-pypi-and-release-new-version-using-github-actions-108k) on how to do this with GitHub Actions.

For now, we will publish **only pre-releases**, with versions ending in `.dev0`, `.dev1`, etc.

Documentation for end users should be updated to reflect this change, in particular, this page: https://muse-os.readthedocs.io/en/latest/installing-muse.html

dalonsoa:
Fixed in bc3a249

PR #123, opened 2023-05-23
Develop imperial

dalonsoa:
# Description

Combines the development done by the RSE team in different fronts and individually described in:

- #115 
- #118 
- #119 
- #121 

Mostly they refer to modernising the infrastructure of MUSE and taking the steps to easy its distribution. A couple of extra issues were open as a consequence of this. See #120 , #109 

## Type of change

## Key checklist

- [x] The documentation builds and looks OK: `$ python setup.py build_sphinx`

## Further checks

### DIFF
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -18,4 +18,4 @@ jobs:
         uses: ruzickap/action-my-broken-link-checker@v2
         with:
           url: https://muse-os.readthedocs.io/en/latest/
-          cmd_params: '-v --buffer-size=8192 --max-connections=3 --color=always --skip-tls-verification --header="User-Agent:curl/7.54.0"'  # muffet parameters
+          cmd_params: '--timeout=20 --buffer-size=8192 --max-connections=3 --color=always --skip-tls-verification --header="User-Agent:curl/7.54.0"'  # muffet parameters
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -1,0 +1,19 @@
+on:
+  schedule:
+    - cron: 0 0 1 * * # run monthly
+  repository_dispatch: # run manually
+    types: [check-link]
+  push:
+    branches: [main, develop_imperial]
+
+name: Broken Link Check
+jobs:
+  check:
+    name: Broken Link Check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Broken Link Check
+        uses: ruzickap/action-my-broken-link-checker@v2
+        with:
+          url: https://muse-os.readthedocs.io/en/latest/
+          cmd_params: '--buffer-size=8192 --max-connections=3 --color=always --skip-tls-verification --header="User-Agent:curl/7.54.0"'  # muffet parameters
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -4,7 +4,9 @@ on:
   repository_dispatch: # run manually
     types: [check-link]
   push:
-    branches: [main, develop_imperial]
+    branches: [main, develop]
+  pull_request:
+      types: [opened, synchronize, reopened]
 
 name: Broken Link Check
 jobs:
--- a/.github/workflows/broken-link-check.yml
+++ b/.github/workflows/broken-link-check.yml
@@ -18,4 +18,4 @@ jobs:
         uses: ruzickap/action-my-broken-link-checker@v2
         with:
           url: https://muse-os.readthedocs.io/en/latest/
-          cmd_params: '--buffer-size=8192 --max-connections=3 --color=always --skip-tls-verification --header="User-Agent:curl/7.54.0"'  # muffet parameters
+          cmd_params: '-v --buffer-size=8192 --max-connections=3 --color=always --skip-tls-verification --header="User-Agent:curl/7.54.0"'  # muffet parameters
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
 ]
 description = "Energy System Model"
 readme = "README.md"
-requires-python = ">= 3.7"
+requires-python = ">= 3.8, <3.10"
 keywords = ["energy", "modelling"]
 classifiers = [
     "Development Status :: 3 - Alpha",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,9 +91,6 @@ version = {attr = "muse.VERSION"}
 [tool.pyls]
 configurationsources = ["flake8"]
 
-[tool.aliases]
-test = "pytest"
-
 [tool.pytest.ini_options]
 testpaths = ["tests", "src/muse"]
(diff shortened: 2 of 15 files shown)


.github/workflows/broken-link-check.yml
```diff
@@ -0,0 +1,19 @@
+on:
+  schedule:
+    - cron: 0 0 1 * * # run monthly
+  repository_dispatch: # run manually
+    types: [check-link]
+  push:
+    branches: [main, develop_imperial]
```
dalonsoa, line 7:
```suggestion
    branches: [main, develop]
```

setup.py
sgiarols, line 1: python setup.py build_sphinx
is not working anymore

sgiarols (changes requested):
It looks good. I have a few points to comment on
- the command on creating the documentation is not working anymore nor it does the command on the PR template python setup.py build_sphinx
- the README file should be updated on the use of the "muse" command rather than reporting just python -m muse ...
- fullsim[trade] fails on my laptop (MACOS, python version 3.9.0). I could see that the server does not use xlrd-1.2.0 (which I have) whereas the server environment uses more packages (
alabaster-0.7.13 ;
babel-2.12.1 ;
certifi-2023.5.7 ;
charset-normalizer-3.1.0 ;
docutils-0.18.1 ;
et-xmlfile-1.1.0 ;
imagesize-1.4.1 ;
latexcodec-2.0.1 ;
markdown-it-py-2.2.0 ;
mdit-py-plugins-0.3.5 ;
mdurl-0.1.2 ;
myst-parser-1.0.0 ;
nbsphinx-0.9.1 ;
openpyxl-3.1.2 ; 
pandoc-2.3 ;
plumbum-1.8.1 ;
ply-3.11 ;
pybtex-0.24.0 ;
pybtex-docutils-1.0.2 ;
requests-2.31.0 ;
snowballstemmer-2.2.0 ;
sphinx-5.0.2 ;
sphinxcontrib-applehelp-1.0.4 ;
sphinxcontrib-bibtex-2.5.0 ;
sphinxcontrib-devhelp-1.0.2 ;
sphinxcontrib-htmlhelp-2.0.1 ;
sphinxcontrib-jsmath-1.0.1 ;
sphinxcontrib-qthelp-1.0.3 ;
sphinxcontrib-serializinghtml-1.1.5 ;
urllib3-2.0.2 )




Regarding the broken links in the documentation, there are many which do not seem real errors. For example:
- time out messages refer to external websites, they are not real errors. 
- most of "not found" items where internal links to pages, not real errors
However, errors "404" were actual ones. For example this one https://github.com/SGIModel/MUSE_OS/blob/main/docs/tutorial-code/5-add-service-demand/1-exogenous-demand/technodata/preset/Residential2050Consumption.csv
It reminds to link to data files used for the tutorial. It bring back again the question on how far we can get the checks on this tutorial running after the model is updated

Shall we tweak a bit this error function to reduce the errors to the actual ones before updating the documentation?

pyproject.toml
```diff
@@ -2,4 +2,128 @@
...
+authors = [
+    {name = "Sustainable Gas Institute", email = "sgi@imperial.ac.uk"},
+]
+description = "Energy System Model"
+readme = "README.md"
+requires-python = ">= 3.8, <3.10"
+keywords = ["energy", "modelling"]
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Programming Language :: Python :: 3.8",
```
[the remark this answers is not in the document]
dalonsoa: Correct. Amended in next commits.

dalonsoa:
> fullsim[trade] fails on my laptop (MACOS, python version 3.9.0). I could see that the server does not use xlrd-1.2.0 (which I have) whereas the server environment uses more packages...

@sgiarols , could you open an issue explaining how you have run that model, error message and package versions you are using. Otherwise it is rather difficult to isolate and figure out the problem.

dalonsoa:
Building the documentation has been fixed in 4cf01e89

The old way of doing it, `python setup.py build_sphinx`, is deprecated and Sphinx itself does not support static configuration files - it was able to use it before via `setup.py`. That's not really an issue as all the information was already in `docs/conf.py`, so it was redundant, anyway. 

I have updated the README and the pull request template to indicate the correct way of building the documentation.

dalonsoa:
About the broken links:

- If a link to an external website fails because that website is not working, it might be just a glitch and it will work again at some point or it might be truly that the URL to the external website is no longer correct. Then is a problem in MUSE documentation and that link should be updated. For example `https://www.anaconda.com/what-is-anaconda/` really does not exist and needs to be updated. Also, for papers you should use proper DOIs rather than URLs, so https://www.sciencedirect.com/science/article/pii/S0306261920308072 should be included as https://doi.org/10.1016/j.apenergy.2020.115295, which is guaranteed to exist (by the definition of DOI, as far as I can tell).
- The internal links are also related to locations that do not exist. For example, if you go to https://muse-os.readthedocs.io/en/latest/source/muse.outputs.html#muse.outputs.YearlyAggregate, it indeed open the page, but does not go to the tag `#muse.outputs.YearlyAggregate`, as required, because that tag does not exist. What does exist is `#muse.outputs.sinks.YearlyAggregate`. I guess that at some point the location of that class was changed, but the documentation was not. And the same with the others.

So, I don't see why they should not be flagged as errors. 

About the errors related to missing files I'm not totally sure what you mean. Could you elaborate?

I'm going to investigate the options of the tool to see if it can tell us where the broken link is located, not just that there is a broken link, which will make sorting them out much easier.

dalonsoa:
It is not possible to report the location of the broken link, at the moment.[ I've created an issue in their repo](https://github.com/ruzickap/action-my-broken-link-checker/issues/104) to see if that can be added. For now, we need to review the documentation manually - using the `search` function in VSCode, obviously.

dalonsoa:
About updating the way of using MUSE in the readme, I will do it once we have finish #106 , so we provide an even simpler way of doing things. For now, the existing instructions are still valid.

dalonsoa:
@sgiarols , I've added the automatic deployment of MUSE to PyPI. It has taken a few attempts to get the configuration right, but it is working, now. The instructions in the README will not work until we push a new tag to PyPI - for now, [it is only in TestPyPI](https://test.pypi.org/project/MUSE-OS/), which is not fit for production, only for testing, as the name suggests. 

I will clean the history and clear the tags once you have reviewed and approved the PR - but before merging - and we will make a proper release as MUSE 1.1.0 once this PR is merged.

sgiarols (approved):

dalonsoa:
@sgiarols , I've clean the history and removed the unnecessary tags, so this is ready for merging.

sgiarols:
> > fullsim[trade] fails on my laptop (MACOS, python version 3.9.0). I could see that the server does not use xlrd-1.2.0 (which I have) whereas the server environment uses more packages...
> 
> @sgiarols , could you open an issue explaining how you have run that model, error message and package versions you are using. Otherwise it is rather difficult to isolate and figure out the problem.

done here #127

PR #149, opened 2023-08-05
Description bisection rule for carbon budget

### LINKED ISSUE #143: Carbon budget control and assumptions need more detailed, dedicated description
## Is your feature request related to a problem? Please describe.
The control of the simulation based on the carbon budget is a very cool feature and very much a hot topic. However, the capabilities of MUSE in that respect are poorly documented, as well as the underlying assumptions. At the moment, there's only some description when describing the [simulation settings](https://muse-os.readthedocs.io/en/flow_charts/inputs/toml.html#carbon-market), but that feels rather hidden and does not go into the detail on the algorithms behind the feature. 

More details of the feature in a dedicated section featuring prominently, so it is clear it exists, its purpose and what it does. 

Blindly copying example input files, and try to figure out what is going on based on the scattered documentation.

## Additional context
None

sgiarols:
# Description

*This PR adds comments on relevant sections of the code for carbon budget estimates based on the bisection rule.*

Fixes #143 

## Type of change

## Key checklist

## Further checks


src/muse/carbon_budget.py
```diff
@@ -336,14 +336,30 @@ def bisection(
     price_too_high_threshold: float = 10,
     fitter: Text = "slinear",
 ) -> float:
-    # to be set as moving value in superloop before emissions start increasing
+    """Applies bisection algorithm to escalate carbon price and meet the budget.
```
dalonsoa, line 339: I feel this needs a more detailed description of what this algorithm is about. Probably, also the other algorithm needs it, but given that we are clarifying this one, let's make it right.

dalonsoa (changes requested):
I feel the docstrings could be a bit more informative, and possibly the code could be broken down into smaller bits. Otherwise, it looks good. 

I wonder why not using an[ out of the box bisection solver](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.bisect.html). Granted it could be tricky to setup, considering the complexity of the function to optimise, but we wouldn't need to worry about setting up the boundaries of the interval with each iteration.

dalonsoa:
Also, I don't think this PR should close #143 - that one is more broadly about how the Carbon Budget is described in the documentation, and not just about the bisection method, which is only part of what needs documenting.

dalonsoa:
HI @sgiarols , it would be good to take this out of the way at some point. Could you look into my comments and suggestions?

sgiarols:
I addressed the changes requested about improving documentation and the notes to the code

dalonsoa (approved):

PR #152, opened 2023-08-08
Builds standalone, GUI-based versions of MUSE

### LINKED ISSUE #150: Build standalone, MUSE executable
## Is your feature request related to a problem? Please describe.
Despite the effort to make the installation of MUSE as simple and clear as possible, it is still a lot of effort for end users that only want to start using MUSE as fast and painless as possible. In the best case scenario, at least a compatible version of Python needs to be installed, which might be painful.

One possibility would be to deploy MUSE as a standalone executable (a `.exe` in Windows or an `app` in MacOS) using [`pyinstaller`](https://pyinstaller.org/en/stable/), a tool the RSE team is familiar with. Those executables bundle python within them and so they are self-contained and standalone. If added to the path, MUSE could be executed from anywhere in the OS.

The main caveat of this is that they cannot be upgraded when a new version of MUSE is release: the user will need to delete the old version and download and install the new one.

Installing MUSE the usual way. 

## Additional context
None

### LINKED ISSUE #151: Add a simple GUI to MUSE
## Is your feature request related to a problem? Please describe.
Some users will not be comfortable with the terminal, so having a GUI will be convenient to ease the interaction with MUSE. 

Developing a full-feature GUI is a daunting task that requires dedicated effort. However, a minimal front end can be added to MUSE easily exploiting the fact it uses `click` to parse the input arguments by making use of [`cli2gui`](https://github.com/FHPythonUtils/Cli2Gui). If we combine this with #150 , it will mean having a standalone, GUI-based version of MUSE. Probably it won't be the solution for complex, long simulations, but it should serve most users most of the time. 

None. A GUI has not been considered, so far. 

## Additional context
None

dalonsoa:
# Description

This is the first attempt to build standalone versions of MUSE that do not require installing anything beforehand - just download the binary -, as well as including a minimal GUI to easy the interaction with the tool. 

Fixes #150 
Fixes #151 

## Type of change

## Key checklist

## Further checks
N/A


sgiarols (commented):
It would be good to have the documentation updated on the existence of GUI as well as to how to install it from the repository,

dalonsoa:
I'll was planning to, but it will be good to have some feedback on that first.

sgiarols:
I cannot get the GUI working, always getting stuck at this error. I installed as pip install -e . [dev,gui] on WIndows PS
image

I edited Environment Variable to point where the scripts should be, but probably something else is missing?

dalonsoa:
I think you are mixing instructions. Adding the location of the `.exe` to the path is only necessary if you are downloading the standalone version of MUSE. If that were the case, I think it should work with no problems under Windows. You do not need to install anything in this case: just download and unzip. That's all. And add to the path, optionally.

If you are installing the source version with `pip`, there are is no folder with `.exe` to add to the path (unless you run `pyinstaller`, which I don't think you're doing). What are you adding to the path, in this case?

If you install MUSE with `pip`, as you are doing, then MUSE should run the usual way: `python -m muse ...` or `python -m muse_gui`, assuming you have the environment activated, as usual. Probably it should also run if you use directly `muse ...` and `muse_gui`, but I have to admit I have not checked that.  

So, what specific steps are you following?

sgiarols:
A first problem is that the standalone version is not downloadable from the repository, as the corresponding tests do not run. Therefore the installation is done with pip, as mentioned earlier.

From the command line, I ran muse just typing "muse" plus the path to settings.toml; I ran the gui typing "muse_gui". The first works, the second does not.

dalonsoa:
I've been checking this and I can reproduce the error, but I cannot figure out what's wrong. For now:

This works:
- `muse`
- `python -m muse`
- `python -m muse_gui`
- `muse.exe` (if using standalone version)
- `muse_gui.exe` (if using standalone version)

This does not work:
- `muse_gui`

I'm going to open an issue about it, but I'm about to go on annual leave, so it might take a while for me to fix it.

sgiarols (commented):
I installed muse and muse_gui  using pyinstaller. Some notes below; some could be addressed in this PR on next ones, as agreed.

- it seems muse and muse_gui cannot be open simultaneously, is it a desired feature?
- I cannot run muse.exe from dist/MUSE in my current configuration, what happens is that the screen of a terminal does come up for 2 seconds and disappears. What is strange is that the window of the terminal is similar to the anaconda prompt. Can it be a conflict with previous anaconda installation? However, why should there be a link with anaconda prompt?
- Regarding the installation for developers, it is worth adding to the documentation the need to activate an environment. Otherwise, upon clicking on muse_gui.exe, we may receive errors on missing packages such as Goosey and we cannot open muse_gui.
- Regarding the interface, I wonder whether the button names can be changed. Now "edit" allows to set a new simulation Maybe we could use "New Simulation" (rather than the current "edit") and "Restart Simulation" (rather than the current "restart")
- Can we have a button that clears all the fields?
- Can we have a button that when starting a new simulation clears the Terminal?
-  Regarding the interface, is it useful to give both the option to select the TOML or to select the working directory? If we just choose the second way, if a TOML is named differently from settings.TOML there is an error. Also, selecting the directory, does notinform the location of the TOML file. For example, as shown in the screenshot, the wdir is selected but the path to TOML is by default to the last MUSE call. I would keep only the first option of defining the path to TOML unless the wdir selection can be used to inform the TOML. location. Incidentally, if keeping both the  "path-to-TOML" and wdir, it may be worth having the working directory selected first and then the TOML.
- It could be made clearer which model I am running probably raising an error or warning if I have conflicting selection. In the second screenshot I have the default_timeslices in the required field but the default model in the optional. The default model runs at the end.

Screenshot 2023-08-16 at 10 47 16
Screenshot 2023-08-16 at 11 04 03

dalonsoa:
I'm not sure I fully understand your points. I'll try to answer below:

> * it seems that muse and muse_gui cannot be open simultaneously, is it a desired feature?

Why do you say so? I can run multiple instances of the GUI and the command line tool simultaneously with no problem. The only trick might be if they are running the same model and try to access the very same input/output files simultaneously. What error message are you getting?

> * I cannot run muse.exe from dist/MUSE in my current configuration, what happens is that the screen of a terminal does come up for 2 seconds and disappears. What is strange is that the window of the terminal is similar to the anaconda prompt. Can it be a conflict with previous anaconda installation? However, why should there be a link with anaconda prompt?

`muse.exe` is a command line tool, so it needs to run from the command line. If you just click in the icon what happens is exactly what you describe. If you open a terminal, navigate to dist/MUSE and then run `muse.exe` from there, it will work fine (or from anywhere if you have it in the path). You might need to use `./muse.exe`, depending on how Windows is configured. 

> * Regarding the installation for developers, it is worth adding to the documentation the need to activate an environment. Otherwise, upon clicking on muse_gui.exe, we may receive errors on missing packages such as Goosey and we cannot open muse_gui.

Will do. 

> * Regarding the interface, I wonder whether the button names can be changed. Now "edit" allows to set a new simulation Maybe we could use "New Simulation" (rather than the current "edit") and "Restart Simulation" (rather than the current "restart")
> * Can we have a button that clears all the fields?
> * Can we have a button that when starting a new simulation clears the Terminal?

If we design a bespoke interface, everything can be done. However, what we have now is an automatic GUI generated out of the command line options, so there is no flexibility there. What you see is what you get with the minimal effort invested.

> * Regarding the interface, is it useful to give both the option to select the TOML or to select the working directory? If we just choose the second way, if a TOML is named differently from settings.TOML there is an error. Also, selecting the directory, does notinform the location of the TOML file. For example, as shown in the screenshot, the wdir is selected but the path to TOML is by default to the last MUSE call. I would keep only the first option of defining the path to TOML unless the wdir selection can be used to inform the TOML. location. Incidentally, if keeping both the  "path-to-TOML" and wdir, it may be worth having the working directory selected first and then the TOML.

The working directory has nothing to do with where the TOML is located. It informs python what is the working directory such that the paths in the TOML file that depend on that location (those that have `{cwd}` in them) can be updated appropriately. Check the last part of this page of the documentation: https://muse-os.readthedocs.io/en/latest/inputs/toml_primer.html . As you can see, that is a different place than `{path}`, which indicates where the TOML file is located. 

For the command line muse this didn't matter much as your working directory is the one from where you run the command. There is no ambiguity. But for the GUI, that's problematic as the working directory is generally uncertain. This option in the GUI is precisely to let the user define that working directory.  

> * It could be made clearer which model I am running probably raising an error or warning if I have conflicting selection. In the second screenshot I have the default_timeslices in the required field but the default model in the optional. The default model runs at the end.

This is exactly the same thing that happens with the command line. The GUI just reproduces its behavior. Under the hood, they are both running [the same function](https://github.com/SGIModel/MUSE_OS/blob/906d6f8620fff5c41abea9b878a2d3a09d0dc527/src/muse/__main__.py#L30) that decides what to run. As you can see, providing a model takes precedence over running a user provided TOML file. We can change that behavior, but just to be clear, it is the one that existed so far, I have not changed it.

sgiarols (approved):
At this stage, on my Windows, muse.exe does not open upon double-clicking. However since all the steps should be properly documented, I will re-check, repeating the steps, after the documentation will have been updated.
I can approve for now, assuming that the documentation will clarify the installation methods and the usage of the interface (which are points raised in the PR review).

dalonsoa:
Ok, but I still do not understand your problem. `muse.exe` is a command line tool so what do you expect to happen when you double click on it beyond, maybe, a terminal that opens and closes immediately?

sgiarols:
Exactly the point is that nothing else happens after the terminal quits. I would have expected the be asked about which simulation to run and see the simulation progressing, as you previously showed in your video.
Let's re-check once the documentation is provided.

dalonsoa:
@sgiarols , please check the video carefully: in there I click `muse_gui.exe` and NOT `muse.exe`.

PR #167, opened 2023-09-07
Remove StarMUSE from docs

### LINKED ISSUE #163: Mention of StarMuse should be replaced by MUSE in the documentation
Here and there in the documentation one can see mentions of StarMuse, referring to the old, global model. That should be replace by MUSE everywhere.

dalonsoa:
# Description

Mainly, this PR removes the mention of StarMUSE in the documentation. While not a big deal, it might be confusing to users. 

As a bonus, I've added `seaborn` as dependency for the docs as the notebooks it contains fail to run unless that dependency is available. In general, it seems to be that ALL notebooks used in the documentation should be re-run just before building the documentation to ensure it contains up to date information (and that they actually run). I'll open an issue about that. 

As a second bonus, I've made a small change in one of the notebooks as it was using the full path in Alex's system to load data. I've change it to relative paths (and use the `pathlib` instead of string-based paths, which is a more modern approach).

Fixes #163 

## Type of change

- [x] Update to the Documentation (non-breaking change)

## Key checklist

## Further checks

N/A


docs/tutorial-code/4-modify-timing-data/2-modify-time-framework/visualise-time-framework-changes.ipynb
```diff
@@ -13,8 +18,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
-   "metadata": {},
+   "execution_count": 6,
+   "metadata": {
+    "ExecuteTime": {
```
sgiarols, line 23: Why do we have the execution time?

sgiarols (approved):
I would agree that the documentation should build the case study on the go, any time is being built.

PR #175, opened 2023-10-04
Run examples in MUSE as regression tests

### LINKED ISSUE #161: Run existing examples as regression tests
MUSE has a collection of examples that could be run automatically as regression tests. This will allow to easily identify changes that break full scale simulations - or that results in numerically different results. 

These regression tests should be run automatically (every week, for example, to monitor they are all passing with new updates in dependency versions) as well as part of the normal testing suite with every PR to main or develop - unless they are really, really time consuming. 

This is partly done, already, but some examples are missing: https://github.com/SGIModel/MUSE_OS/blob/f89eb147a6a706fa8f180717f7eefd8336662d0c/tests/test_fullsim_regression.py#L7

dalonsoa:
# Description

This was partly implemented, but not in all the examples. Now it is and all the example will run as regression tests (comparing the output with the expected real output) with every run of the CI system and also periodically, every 2 weeks.

The expected real output has been moved to the tests folder as that's a more appropriate place to be than with the MUSE code.

The CI is currently failing precisely because two of this examples are failing to run: medium and multiple_agents. See errors below. 

```
FAILED tests/test_fullsim_regression.py::test_fullsim_regression[medium] - muse.errors.GrowthOfCapacityTooConstrained: Error during the investment process. The capacity was not allowed to grow
FAILED tests/test_fullsim_regression.py::test_fullsim_regression[multiple_agents] - ValueError: technology must not be empty
```

You can reproduce simply by trying to run the examples manually:

```
python -m muse --model medium
python -m muse --model multiple_agents
```

@sgiarols , @martinstringer , I've open this as a draft PR as it makes no sense to merge it when this is failing. Considering that this is what we tell people to use for learning and to start their own projects, it will be good to figure out what's wrong as soon as possible.

Fixes #161 

## Type of change

## Key checklist

## Further checks


sgiarols:
@dalonsoa, let's check first whether we can merge #182 and then I will review this one

dalonsoa:
@sgiarols , the changes you made in #182 affect #177 but not this one. This is still failing with the same complains.

sgiarols (approved):

sgiarols:
I now update the outputs

sgiarols (commented):
It seems there are still some issues in the tests, despite those being running and working offline.

sgiarols:
Now the results from examples should be in the right folder

sgiarols:
@dalonsoa, I think that as for the settings.TOML results are sento to different folders?

dalonsoa:
I think the data was in the wrong place, so the comparison was failing. I've re-run the example locally and check that, indeed, the output tree was different. I've updated the "true" results in `tests/example_outputs/multiple_agents` with the right information obtained from a local run, and now the regression tests seems to be passing. 

One of the unit tests is now failing, though. It might be using one of the examples you modified and, as a results, if you changed the units, now the actual numbers produced are wrong.

dalonsoa:
Tracing down the error, it comes down because something was changed in the configuration for the `medium` example, so the numerical values no longer match the expectation. If you're sure the example is working well, the solution should be as simple as replacing the old (wrong) expected values with the current ones in the 2 failing tests.

sgiarols:
Thanks. We are nearly there with this PR. I only have this error from codecov. I thought you would better know what this means and how to fix this.

dalonsoa:
It's not an error, just Codecov being picky. Sorted in https://github.com/SGIModel/MUSE_OS/pull/186 . I'll merge this as it is prefectly fine from our purposes.

PR #247, opened 2024-03-02
Variable costs definition

### LINKED ISSUE #223: Incorrect variable cost equation
## Describe the bug
The equation in the documentation section 9.3.3 that describes variable cost is incorrect. 

The application of the variables var_par and var_exp within the model also appears to be inconsistent...

In the calculation of net present value and levelled cost in objectives.py, it seems to be proportional to production**var_exp, for example on line 820:

`variable_costs = var_par *((production.sel(commodity=products).sum("commodity")) ** var_exp`

However for the calculation of gross margin in quantities.py (line 170) it seems to be proportional to capacity**var_exp:

`variable_costs = convert_timeslice(var_par * capacity**var_exp, prices.timeslice, QuantityType.EXTENSIVE)`

## Screenshots
Screenshot 2023-12-12 at 19 44 31

sgiarols:
# Description

*Variable costs are dependent on capacity ib the gross margin definition, rather than being proportional to the capacity.*

Fixes #223 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/quantities.py
+++ b/src/muse/quantities.py
@@ -165,16 +165,18 @@ def gross_margin(
     fixed_outputs = tech.fixed_outputs
     fixed_inputs = tech.fixed_inputs
 
-    # Hours ratio
-    variable_costs = convert_timeslice(
-        var_par * capacity**var_exp, prices.timeslice, QuantityType.EXTENSIVE
-    )
-
     prices = prices.sel(region=capacity.region).interp(year=capacity.year)
 
     # Filters
     environmentals = is_pollutant(technologies.comm_usage)
     enduses = is_enduse(technologies.comm_usage)
+
+    # Hours ratio
+    variable_costs = convert_timeslice(
+        var_par * ((fixed_outputs.sel(commodity=enduses)).sum("commodity")) ** var_exp,
+        prices.timeslice,
+        QuantityType.EXTENSIVE,
+    )
 
     # The individual prices
     consumption_costs = (prices * fixed_inputs).sum("commodity")
--- a/src/muse/quantities.py
+++ b/src/muse/quantities.py
@@ -132,8 +132,11 @@ def emission(production: xr.DataArray, fixed_outputs: xr.DataArray):
 def gross_margin(
     technologies: xr.Dataset, capacity: xr.DataArray, prices: xr.Dataset
 ) -> xr.DataArray:
-    """profit of increasing the production by one unit.
-
+    """The percentage of revenue after direct expenses have been subtracted
+    .. _reference:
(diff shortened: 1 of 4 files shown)


src/muse/quantities.py
```diff
@@ -164,18 +164,25 @@ def gross_margin(
...
+    else:
+        prices = prices.where(prices.region == capacity.region, drop=True)
+    prices = prices.interp(year=capacity.year.values)
 
     # Filters
     environmentals = is_pollutant(technologies.comm_usage)
     enduses = is_enduse(technologies.comm_usage)
 
+    # Hours ratio
+    variable_costs = convert_timeslice(
```
dalonsoa, line 180: It migth be worth including in the docstring the equation that this represents, so the code is easier to understand. Even better if a reference to a paper or web is included giving more context.

tests/test_quantities.py
```diff
@@ -104,12 +105,18 @@ def test_gross_margin(technologies, capacity, market):
...
     market.prices[{"commodity": is_pollutant(usage)}] = env_prices = 6
     # We expect a xr.DataArray with 1 replacement technology
-    actual = gross_margin(technologies, capacity, market.prices)
+    actual = gross_margin(technologies, capa, market.prices)
 
     revenues = prices * prod * sum(is_enduse(usage))
     env_costs = env_prices * envs * sum(is_pollutant(usage))
     cons_costs = prices * fuels * sum(is_fuel(usage))
-    var_costs = vp * (capa**ve) * market.represent_hours / sum(market.represent_hours)
+    var_costs = (
```
dalonsoa, line 113: As above, it will be clearer if some context is added to the docstrings.

dalonsoa (approved):
I've just suggested to update the doscstrings to explain in more detail what's going on. Assuming that the equations are ok, the code looks good to me.

tests/test_quantities.py
sgiarols: Thanks. I added a link and an explanation of the terms.
Also, it is more common to express the gross margin as a percentage of the revenues. Therefore I added this additional calculation.

src/muse/quantities.py
sgiarols: Thanks. I added the links and added explanations.
One thing I corrected, is that it is more common to define the gross margin as a percentage of the revenues. Therefore I added these extra steps.

martinstringer (approved):
These look consistent as far as I can tell in the time available.

martinstringer:
Thank you Sara, we were not expecting you to make these changes yourself necessarily - just to confirm we were correct in identifying the issue - but thank you for going ahead and addressing it.

Issue #252, opened 2024-03-11
Error message caused by demand sharing with multiple agents [BUG] 


LennartMorlock:
## Describe the bug
When running MUSE with multiple agents with different objectives, search rules or decision methods, the error message below often occurs. This also relates to a [post in the MUSE Google group](https://groups.google.com/g/muse-model/c/YyfVv46C65U/m/YGUuFpdFAQAJ).

Traceback (most recent call last):
  File "C:\Users\lenna\anaconda3\envs\muse_env\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\lenna\anaconda3\envs\muse_env\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\lenna\anaconda3\envs\muse_env\Scripts\muse.exe\__main__.py", line 7, in <module>
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\__main__.py", line 60, in run
    muse_main(args.settings, args.model, args.copy)
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\__main__.py", line 55, in muse_main
    MCA.factory(settings).run()
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\mca.py", line 346, in run
    _, new_market, self.sectors = self.find_equilibrium(new_market)
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\mca.py", line 209, in find_equilibrium
    return find_equilibrium(
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\mca.py", line 561, in find_equilibrium
    market, equilibrium_sectors = single_year_iteration(market, sectors)
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\mca.py", line 468, in single_year_iteration
    sector_market = sector.next(
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\sectors\sector.py", line 236, in next
    subsector.invest(
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\sectors\subsector.py", line 74, in invest
    lp_problem = self.aggregate_lp(
  File "C:\Users\lenna\MUSE_test2\MUSE_OS\src\muse\sectors\subsector.py", line 147, in aggregate_lp
    result = agent.next(
[...]

PR #253, opened 2024-03-15
Update demand_share.py _inner_split function to solve reoccuring error message when running MUSE with multiple agents

LennartMorlock:
# Description

This commit includes the minor change to the _inner_split function in the demand_share.py file, described in the [related issue](https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/252#issue-2179280270). 

It solves the error. However, uncertainty about its sensibility remains from a modelling perspective.

## Type of change

## Key checklist

## Further checks


martinstringer:
Hi Lennart, thank you from all of us for proposing this fix. If time allows, would you be able to also share your understanding of what your proposed changes are doing? If this could be added directly as comments to the new lines of code, even better.

LennartMorlock:
Sorry for not responding earlier. I created a much-simplified model (attached) that causes the same error aimed to be solved with this pull request. I added it to the related issue as well. 

[simple_model_error.zip](https://github.com/EnergySystemsModellingLab/MUSE_OS/files/15306234/simple_model_error.zip)

alexdewar:
Hi @LennartMorlock. Thanks for your reply. I was able to replicate the issue with the config files you attached.

I'm not sure what this function is supposed to be doing either tbh! It's a little opaque. I'm tagging my colleague @dalonsoa who might have a better idea.

dalonsoa:
Not much to add to what is said in the docstring. The demand of any commodity needs to be split between the agents so they can use their share of the demand to invest in new technologies. Now, why there's an `unassigned`, why is calculated that way and its relationship with `total`, I don't really know.

alexdewar:
Okey doke. I'll have a play with it a bit more to see if I can understand it better, but provided that it passes tests, would you be happy if this is merged?

dalonsoa:
After updating the branch, yes, I think that's fine.

alexdewar (approved):
This looks godo to me @LennartMorlock. I wasn't entirely able to figure out what was happening with that `unassigned` variable, but your change seems sensible.

I've had a go at tidying it up a bit. I don't think we need to mention the old code in the comments. Also, I extracted the sum of the `shares` objects into its own variable for the sake of readability and not performing the calculation twice.

I've rerun it over your sample config and it seems to still produce the same result, but would you mind checking that what I've done makes sense to you, please? If you're happy with it I'll merge this branch.

LennartMorlock:
Hi @alexdewar, 
yeah, the changes seem good to me. I still get the same results and the code definitely looks cleaner.

alexdewar:
Great, thanks :smile:

Issue #254, opened 2024-03-26
maturity and spend_limit filters adapt the search_space faulty [BUG] 


LennartMorlock:
## Describe the bug
Some filters (I have noticed that maturity and spend_limit do this, but I am not sure if there is another one) alter the search_space by removing the columns from the search_space DataArray that belong to replacement technologies that don't satisfy the condition and should not be considered. That leads to two problems:

1. Some filters, like spend_limit, require all replacement columns to remain present when applied. Otherwise, applying the .sel() function within the filter may throw an error as not all indices it tries to select can be found in the search_space.

2. with_asset_technology should add the existing asset being replaced to the search space as a possible replacement. It does this by setting the corresponding boolean values to True. However, this requires that the relevant (asset, replacement) combinations exist in the search_space Dataarray, as it does not add a new column with the correct boolean values. Thus, if it is applied after the maturity filter, it can not add the existing technologies to the search_space since those columns don't exist.

## To Reproduce
1. Change an agent's search rule to maturity->spend_limit with high values for maturity and spend_limit such that spend_limit allows  technologies that maturity doesn't

2. Change any retrofit agent's search rule to maturity with a maturity threshold bigger than the highest share of any existing technology. 

## Expected behaviour
1. Should run, however, throws an error message.

2. The retrofit agent should still be able to invest in the existing technologies as the with_asset_technology filter is automatically applied to retrofit agents with one search rule. However, the previously described problem will lead to a critical warning of an empty search space in the log.

## Context
Please complete the following to better understand the system you are using to run MUSE.

- Operative system (eg. Windows 10):  Windows 10
- MUSE version (eg. 1.1.0): 1.1.0
- Installation method (eg. pipx, pip, development mode):  development mode
- Python version (you can get this running `python --version`): 3.9.18

alexdewar:
Would you mind attaching config files which replicate the bug, please?

PR #255, opened 2024-03-26
Update returned search_space of maturity and spend_limit filter

LennartMorlock:
## Description

Two new lines of code were added to the maturity and the spend_limit filters, and one line was replaced with them. The changes to both filters are identical. 

The previous replaced line removed the columns corresponding to replacements that should not be considered according to the search rule. The two new lines turn the corresponding boolean values inside the search space Dataarray to False instead while keeping all columns. 

This solves the related issue #254 , as it removes the possibility of an error if the filters are put in a row (e.g. maturity->spend_limit) and makes the with_asset_technology filter work normally if it is applied afterwards.

Finally, this makes the maturity and spend_limit filters also more in line with other filters like same_fuel or similar_technology in how they manipulate the search space as they also just change the boolean values of the search_space DataArray while keeping the dimensions identical.

Fixes # Issue 254

## Type of change

## Key checklist

## Further checks


alexdewar:
I'm merging in `develop` to check that the tests still pass.

alexdewar:
This seems sensible enough.

alexdewar (approved):

PR #266, opened 2024-04-18
Update example output

### LINKED ISSUE #258: Example output in documentation is broken [BUG]
## Describe the bug

The `example-output.ipynb` notebook (in `docs`) is trying to run MUSE with a settings file that doesn't exist. Hence, rather than showing a useful example of a MUSE output, it's just showing an error message that `settings.toml` doesn't exist.

Either a settings file needs to be added, or the notebook needs to be modified to point to an existing file, or run the default simulation

## To Reproduce
The output is shown [here](https://muse-os.readthedocs.io/en/latest/example-output.html) (and can also be reproduced by building the documentation locally)

tsmbland:
# Description

this change.*

Fixes #258 

## Type of change

## Key checklist

## Further checks


alexdewar (approved):
LGTM!

Side note: @dalonsoa Is there a reason why results files for the examples are committed to the repo? It seems like the sort of thing that would normally be `.gitignore`'d. Though I can see that we might want sample outputs for users to plot etc...

dalonsoa:
> LGTM!
> 
> Side note: @dalonsoa Is there a reason why results files for the examples are committed to the repo? It seems like the sort of thing that would normally be `.gitignore`'d. Though I can see that we might want sample outputs for users to plot etc...

They are used to create the plots in the examples used in the documentations, I think.

dalonsoa:
Mmmmm... maybe you're right, yes. To be honest, I'm not entirely sure if the notebooks are run when the documentation is built in ReadTheDocs.

alexdewar:
I'll check and fix up if needed.

PR #278, opened 2024-04-23
Add `pre-commit` hooks to format config files

alexdewar:
Apologies about all the whitespace-related PRs!

I thought while I was at it, it might be useful to have consistent formatting for our config files too (in particular, there are a lot of TOML files in this repo). I've added the `prettier` hook that we've used elsewhere for JSON and YAML files and another one for TOML (which is sadly not supported by `prettier`).

One upside of doing this is that it will also detect syntax errors in these files.

This is the last thing I wanted to do before adding that `ruff` hook.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,12 +1,21 @@
-# EditorConfig is awesome: http://EditorConfig.org
-
-# top-most EditorConfig file
 root = true
 
-[*.py]
+[*]
+charset = utf-8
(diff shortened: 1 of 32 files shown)


dalonsoa (approved):
What you really want is to appear as contributor to every single file in MUSE xD 

LGTM

tsmbland (approved):
Looks good!

PR #280, opened 2024-04-23
Add ruff `pre-commit` hooks

### LINKED ISSUE #267: Use ruff pre-commit hook
We can use ruff/ruff-format to replace the functionality of isort, black and flake8

alexdewar:
This PR replaces various Python tools with `ruff` in the `pre-commit` hooks. I've tried to keep the list of checks as close to the ones that we were doing before with Flake 8. There were a few small things that needed fixing up, but nothing major. There are some other useful checks ruff can do (e.g. the "pyupgrade" ones), but I think those should be separate PRs so they can be discussed on their merits.

It seems like `ruff` will happily sort out Jupyter Notebooks for you too (both with linting and formatting), so I've added this.

Fixes #267.

Next I have a PR to run `codespell` over the code base, at which point I think I will have touched every file 🙃


tsmbland (approved):
Good stuff!

dalonsoa (approved):
This is brilliant!

PR #289, opened 2024-05-03
WIP: Upgrade and un-pin dependencies and Python version to use

### LINKED ISSUE #120: Revise the version of Pandas being targeted

dalonsoa:
# Description

Update the whole codebase and tests to enable MUSE running with more modern Python versions, `pandas` and `xarray`. So far, due to the versions used of `pandas` and `xarray` we were limited to using Python 3.9, at most. 

Things to do:

- [x] Unpin dependencies in `pyproject.toml`
- [x] Add modern python to workflow files
- [x] Review the code to avoid errors
- [ ] Review the code to ensure tests pass
- [ ] Make sure that notebooks and tutorials still produce the right results (requires researchers intervention)

Fixes #120 

## Type of change

- [x] Breaking change (whatever its nature)

## Key checklist

## Further checks

### DIFF
--- a/src/muse/mca.py
+++ b/src/muse/mca.py
@@ -64,8 +64,12 @@ class MCA(object):
             ).sel(region=settings.regions)
         ).interp(year=settings.time_framework, method=settings.interpolation_mode)
 
-        market["supply"] = zeros_like(market.exports)
-        market["consumption"] = zeros_like(market.exports)
+        market["supply"] = zeros_like(market.exports).drop_vars(
+            ["timeslice", "month", "day", "hour"]
+        )
+        market["consumption"] = zeros_like(market.exports).drop_vars(
+            ["timeslice", "month", "day", "hour"]
+        )
 
         # We create the sectors
         sectors = []
@@ -459,7 +463,9 @@ def single_year_iteration(
     sectors = deepcopy(sectors)
     market = market.copy(deep=True)
     if "updated_prices" not in market.data_vars:
-        market["updated_prices"] = market.prices.copy()
+        market["updated_prices"] = market.prices.copy().drop_vars(
+            ["timeslice", "month", "day", "hour"]
+        )
 
     # eventually, the first market should be one that creates the initial demand
     for sector in sectors:
@@ -547,12 +553,16 @@ def find_equilibrium(
     else:
         included = ones(len(market.commodity), dtype=bool)
 
-    market["updated_prices"] = market.prices.copy()
+    market["updated_prices"] = market.prices.copy().drop_vars(
+        ["timeslice", "month", "day", "hour"]
+    )
     prior_market = market.copy(deep=True)
     converged = False
     equilibrium_sectors = sectors
     for i in range(maxiter):
-        market["prices"] = market.updated_prices
+        market["prices"] = market.updated_prices.drop_vars(
+            ["timeslice", "month", "day", "hour"]
+        )
         prior_market, market = market, prior_market
         market.consumption[:] = 0.0
         market.supply[:] = 0.0
@@ -576,7 +586,7 @@ def find_equilibrium(
             )
             market["prices"] = future_propagation(  # type: ignore
                 market["prices"], new_price
-            )
+            ).drop_vars(["timeslice", "month", "day", "hour"])
 
             break
 
@@ -593,7 +603,7 @@ def find_equilibrium(
             )
             market["prices"] = future_propagation(  # type: ignore
                 market["prices"], new_price
-            )
+            ).drop_vars(["timeslice", "month", "day", "hour"])
         if not equilibrium:
             equilibrium_reached = True
             converged = True
@@ -620,7 +630,7 @@ def find_equilibrium(
         )
         market["prices"] = future_propagation(  # type: ignore
             market["prices"], new_price
-        )
+        ).drop_vars(["timeslice", "month", "day", "hour"])
         getLogger(__name__).critical(msg)
 
     return FindEquilibriumResults(
--- a/src/muse/readers/csv.py
+++ b/src/muse/readers/csv.py
@@ -27,6 +27,21 @@ from muse.defaults import DEFAULT_SECTORS_DIRECTORY
 from muse.errors import UnitsConflictInCommodities
 
 
+def to_numeric(x):
+    """Converts a value to numeric if possible.
+
+    Args:
+        x: The value to convert.
+
+    Returns:
+        The value converted to numeric if possible, otherwise the original value.
+    """
+    try:
+        return pd.to_numeric(x)
+    except ValueError:
+        return x
+
+
 def find_sectors_file(
     filename: Union[Text, Path],
     sector: Optional[Text] = None,
@@ -83,7 +98,7 @@ def read_technodictionary(filename: Union[Text, Path]) -> xr.Dataset:
     data.index.name = "technology"
     data = data.drop(["process_name", "region_name", "time"], axis=1)
 
-    data = data.apply(lambda x: pd.to_numeric(x, errors="ignore"), axis=0)
+    data = data.apply(to_numeric, axis=0)
 
     result = xr.Dataset.from_dataframe(data.sort_index())
     if "fuel" in result.variables:
@@ -128,7 +143,7 @@ def read_technodata_timeslices(filename: Union[Text, Path]) -> xr.Dataset:
     )
     data = csv[csv.technology != "Unit"]
 
-    data = data.apply(lambda x: pd.to_numeric(x, errors="ignore"))
+    data = data.apply(to_numeric)
     data = check_utilization_not_all_zero(data, filename)
 
     ts = pd.MultiIndex.from_frame(
@@ -160,7 +175,6 @@ def read_io_technodata(filename: Union[Text, Path]) -> xr.Dataset:
 
     There are four axes: (technology, region, year, commodity)
     """
-    from functools import partial
 
     from muse.readers import camel_to_snake
 
@@ -180,7 +194,7 @@ def read_io_technodata(filename: Union[Text, Path]) -> xr.Dataset:
     data.columns.name = "commodity"
     data.index.name = "technology"
     data = data.rename(columns=camel_to_snake)
-    data = data.apply(partial(pd.to_numeric, errors="ignore"), axis=0)
+    data = data.apply(to_numeric, axis=0)
 
     fixed_set = xr.Dataset.from_dataframe(data[data.level == "fixed"]).drop_vars(
         "level"
@@ -659,7 +673,9 @@ def read_initial_market(
     result = result.rename(
         commodity_price="prices", units_commodity_price="units_prices"
     )
-    result["prices"] = result["prices"].expand_dims({"timeslice": timeslices})
+    result["prices"] = (
+        result["prices"].expand_dims({"timeslice": timeslices}).drop_vars("timeslice")
+    )
 
     return result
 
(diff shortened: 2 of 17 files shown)


dalonsoa:
There's one notebook test still failing when building the documentation. I'm on it.

alexdewar:
And I've just gifted you a merge conflict :upside_down_face:. Sorry!

alexdewar:
...except my PR broke everything so we'll have to revert it anyway :shrug:

src/muse/readers/toml.py
```diff
@@ -539,17 +539,14 @@ def read_timeslices(
...
-        Data variables:
-            *empty*
+        >>> ts = read_timeslices(toml, ref, transforms)
+        >>> assert "semester" in ts.coords
+        >>> assert "week" in ts.coords
+        >>> assert "day" in ts.coords
+        >>> assert "represent_hours" in ts.coords
+        >>> assert set(ts.coords["day"].data) == {"dusk", "allday"}
+        >>> assert set(ts.coords["week"].data) == {"weekday", "weekend"}
+        >>> assert set(ts.coords["semester"].data) == {"summer", "winter"}
```
alexdewar, line 549: This seems like slightly odd example code. Maybe it would be better as a regular unit test?

src/muse/utilities.py
```diff
@@ -47,8 +47,8 @@ def coords_to_multiindex(
     assert dimension not in data.indexes
     names = [u for u in data.coords if data[u].dims == (dimension,)]
     index = MultiIndex.from_arrays([data[u].values for u in names], names=names)
-    result = data.drop_vars(names)
-    result[dimension] = index
+    mindex_coords = xr.Coordinates.from_pandas_multiindex(index, dimension)
+    result = data.drop_vars(names).assign_coords(mindex_coords)
     return result
```
alexdewar, line 52:
```suggestion
    return data.drop_vars(names).assign_coords(mindex_coords)
```

src/muse/timeslices.py
```diff
@@ -320,27 +321,27 @@ def timeslice_projector(
...
+          * week              (timeslice) object 24B 'weekday' 'weekend' 'weekend'
+          * day               (timeslice) object 24B 'allday' 'dusk' 'night'
 
         It is possible to give as input an array which does not have a timeslice of its
         own:
 
         >>> nots = DataArray([5.0, 1.0, 2.0], dims="a", coords={'a': [1, 2, 3]})
-        >>> timeslice_projector(nots, ref, transforms).T
-        <xarray.DataArray (timeslice: 1, finest_timeslice: 10)>
+        >>> timeslice_projector(nots, ref, transforms).T  # doctest: +SKIP
```
alexdewar, line 337: Out of interest, why are various doctests being skipped? Do you get differences in the floating-point values on different platforms or something?

tests/conftest.py
```diff
@@ -77,9 +77,18 @@ def compare_df(
...
+                expected_col, rel=rtol, abs=atol, nan_ok=equal_nan
+            )
+        except AssertionError:
+            # if the columns are not equal, we check if the sets are equal as sometimes
+            # the order of the rows is different because of different sorting algorithms
+            try:
+                assert set(actual_col) == set(expected_col)
+            except AssertionError:
+                print(f"file: {msg}, column: {col}")
+                raise
```
alexdewar, line 91: Was this debug printing that got left behind?

```suggestion
            assert set(actual_col) == set(expected_col)
```

Besides that, this seems a bit hacky. (Also, given that you're not using `approx()` for the second attempt at comparing, I worry it's maybe a bit fragile.) I'm not against it if it's the only sensible way to do this, but I wonder if we might be able to do something cleaner... Is there no way to sort the rows for `actual_col` and `expected_col` in the same way here, then do the comparison?

alexdewar, line 91: Was this debug printing that got left behind?

```suggestion
            assert set(actual_col) == set(expected_col)
```

Besides that, this seems a bit hacky. (Also, given that you're not using `approx()` for the second attempt at comparing, I worry it's maybe a bit fragile.) I'm not against it if it's the only sensible way to do this, but I wonder if we might be able to do something cleaner... Is there no way to sort the rows for `actual_col` and `expected_col` in the same way here, then do the comparison?

alexdewar (approved):
This looks good! I've tried it with Python 3.12 and it works happily :partying_face: 

I've stuck some minor comments/questions throughout, but on the whole this seems sensible. The repetition of the `drop_vars()` stuff looks a bit clunky to me though I'm not sure if there's a nicer way of doing it... Are we putting columns in the dataframe that shouldn't really be in there?

Besides that, I'm wondering if we need to update the `publish.yml` workflow. I'm guessing that we only need a single wheel for all the Python versions because we don't have any compiled code in here... right? But also we probably want to bump the Python version of the pyinstaller-built binary to 3.12 so it's a bit faster for end-users.

src/muse/readers/csv.py
```diff
@@ -83,7 +98,7 @@ def to_agent_share(name):
     data.index.name = "technology"
     data = data.drop(["process_name", "region_name", "time"], axis=1)
 
-    data = data.apply(lambda x: pd.to_numeric(x, errors="ignore"), axis=0)
+    data = data.apply(to_numeric, axis=0)
```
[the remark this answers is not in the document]
dalonsoa: It used to do it, but not anymore, hence the hack.

src/muse/readers/toml.py
dalonsoa: Possibly. It used to just compare the whole output of the timeslice array, but that is causing trouble as the output used int32 in Windows and int64 in Linux and MacOs, so the comparison in the traditional doctest style was failing.

src/muse/timeslices.py
dalonsoa: Exactly. I do not want to just remove them since they can be illustrative, but as actual examples to run, they are less useful.

src/muse/mca.py
```diff
@@ -64,8 +64,12 @@ def factory(cls, settings: Union[Text, Path, Mapping, Any]) -> MCA:
...
         ).interp(year=settings.time_framework, method=settings.interpolation_mode)
 
-        market["supply"] = zeros_like(market.exports)
-        market["consumption"] = zeros_like(market.exports)
+        market["supply"] = zeros_like(market.exports).drop_vars(
+            ["timeslice", "month", "day", "hour"]
+        )
+        market["consumption"] = zeros_like(market.exports).drop_vars(
+            ["timeslice", "month", "day", "hour"]
+        )
```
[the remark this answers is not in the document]
dalonsoa: We need to drop them because both the `market` dataset and the new `DataArray` we are adding contain those coordinates. In the `old times`, xarray manipulation dealt with that under the hood, but not anymore - at least not for `MultiIndex` coordinates, like `timeslice` - and the conflicting coordinates need to be dropped before assigning them. 

I'm not sure if a helper function would be that helpful. I guess it would be something like:

```python
def drop_timeslice(data: xr.Dataset | xr.DataArray) -> xr.Dataset | xr.DataArray:
    return data.drop_vars(["timeslice", "month", "day", "hour"])
```

Do you think it would be helpful?

tests/conftest.py
dalonsoa: Yes, I was thinking the same thing after implementing this. What do you think of the following?

```python
assert np.sort(actual_col) == approx(
                np.sort(expected_col), rel=rtol, abs=atol, nan_ok=equal_nan
            )
```

src/muse/readers/csv.py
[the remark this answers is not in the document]
dalonsoa: See https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html

src/muse/mca.py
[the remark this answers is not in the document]
alexdewar: I think it would probably be a lot of work for not much gain :stuck_out_tongue:. It just feels a bit messy having the repetition, but I guess that's just to do with the code design and maybe it's just what we have to do.

tests/conftest.py
alexdewar: That looks better :smile:

dalonsoa:
Before merging this we need to decide on the release strategy. The inputs, outputs and functionality are the same (or they should), but we are dropping support for Python 3.8, which might annoy people. 

Also, it will be good if someone with domain knowledge could check the output results of the examples and the tutorials. As I said in the PR description, changes to the output files are very small and, except in a couple of cases, limited to a reordering of the columns or rows, but with the same overall content, so I do not expect non-sensical results, but would be worth checking. @ahawkes , could anyone from your side have a look at this?

ahawkes:
@dalonsoa is there an easy way for me to see the results from the examples and tutorials?

tsmbland:
> @dalonsoa is there an easy way for me to see the results from the examples and tutorials?

I've merged this into my documentation branch, so you see the results from the tutorials [here](https://muse-os.readthedocs.io/en/documentation/index.html) (bearing in mind that there are still issues with tutorials 2 and 7 which are unrelated to these changes). I've given it a quick glance, and as far as I can tell all of the figures look identical compared to how they were before (to see how they were before merging these changes you can download the documentation artifacts from [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/actions/runs/9447521179)).

tsmbland (approved):
As far as I can tell this all looks good! I've tested it with 3.9 and 3.12 at it works fine. The changes to the output files look quite small, and all the graphs in the documentation look the same as they were before.

I have to give some benefit of the doubt as it's a bit too massive to check every line, but I can't see any glaring issues, so I'm happy to approve.

ahawkes:
Yes, this looks good - thanks!

dalonsoa:
Ok, so if everyone is happy we merge and create a release. I'd wait until https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/328 is done and then merge and release first thing tomorrow morning.

ahawkes:
I don't suppose a simple solution is possible to #319 and could be in this release too? Possibly not...

dalonsoa:
I'm in no rush to merge this PR, so happy to wait as long as needed. @alexdewar is on that issue, but I'm not sure of his implementation timeline. I think he was tackling #317 first.

alexdewar:
Yep, I'm working on #317 right now, but I'm hoping to have a chance to look at #319 this week too.

Personally, I don't think we have to issue a new release as soon as we merge this branch. I think it would be ok to merge it, fix these small issues then make the release. The advantage of merging this sooner rather than later is that we'll have a chance to use the code over the next week or so and so there'll be more opportunities to notice bugs before we make a release (although it seems fine).

dalonsoa:
Ok, so I'll merge this tomorrow morning but will hold on the release until #317, #319 and #328 are done, hopefully by the end of the week. Correct?

ahawkes:
Sounds good!

PR #290, opened 2024-05-03
Modify plots in documentation notebooks

### LINKED ISSUE #190: Jupyter notebooks visualisation of the default run is incorrect [BUG]
## Describe the bug
A clear and concise description of what the bug is, including error messages.

Section 4.4 of the documentation is to visualise the results of the default run. It currently produces three plots, each of which is "capacity" in a sector. However, there are several capacity values for each technology in the output file referenced (one vaue for each asset owned by the agent), so the plot method (seaborn lineplot) fails to display the sum of the values. Instead it displays one value with remaining values represented by the "hue". This is wrong.

## To Reproduce
Steps to reproduce the behavior. Attache any input file that might be required.

None required - just run the JN as per the documentation section 4.4.
## Expected behavior

These plots should be bar charts of aggregate capacity in each milestone year. i.e. if there are multiple values in the file they need to be summed and then plotted (as a bar chart, not a line chart).

## Context

- Operative system (eg. Windows 10): Windows 10
- MUSE version (eg. 1.0.1): 1.0.2
- Installation method (eg. pipx, pip, development mode): pipx
- Python version (you can get this running `python --version`): 3.9.13

### LINKED ISSUE #282: Improve figures in tutorial notebooks
## Is your feature request related to a problem? Please describe

Following on from #190, there are similar problems with the figures in the tutorial notebooks

Figures should be changed in a similar way to #277, once that has been approved

tsmbland:
# Description

This PR makes a number of changes to improve the plots in the documentation. The main issue was that plots weren't properly displaying _total_ capacity for each technology. As some of the results files contain multiple capacity values for a given technology in a given year (i.e. multiple assets), and these weren't being summed before plotting, the plots were showing a shaded region to represent the variability across assets, rather than the total (summed) capacity. Main changes:

Modify plots for the default run (Fixes #190):
- I've summed the necessary rows to get aggregate capacity data, and switched the plot type to a stacked bar chart
-  See the original page [here](https://muse-os.readthedocs.io/en/latest/running-muse-example.html#Visualisation), and the new page [here](https://muse-os.readthedocs.io/en/documentation/running-muse-example.html#Visualisation)

Modify plots in the tutorial notebooks (Fixes #282):
- Similar to the above
- For example, compare [this page](https://muse-os.readthedocs.io/en/latest/user-guide/add-solar.html#Running-our-customised-simulation) in the original version, with [this page](https://muse-os.readthedocs.io/en/documentation/user-guide/add-solar.html#Running-our-customised-simulation) in the new documentation
- Similar changes across all the tutorials
- Several other cosmetic changes (e.g. better axis labels)

Notes:
- Some of the text describing the figures may be wrong. I'll come back and check this again at a later date, but I expect that the exact content of the figures may change as I work through the models, so better off fixing the text later
- A also re-ran the models, which led to a change in some of the results files. I'm committing these changes here, however we may eventually want to add these to gitignore (see #287)
- The figures in tutorial 8.2 are wrong (only showing one agent), but this is a problem with the underlying model rather than the figures, so I'll fix that in a separate issue (#291)
- Likewise with 8.7 (issue #295)

## Key checklist

## Further checks


tsmbland:
@ahawkes I've been working through the notebooks to improve the figures. I'm currently using a mix of stacked bar charts and line graphs for the capacity figures (e.g. [this](https://muse-os.readthedocs.io/en/documentation/running-muse-example.html#Visualisation) vs [this](https://muse-os.readthedocs.io/en/documentation/user-guide/add-solar.html#Running-our-customised-simulation)). Should we be more consistent with the type of plot we use, and do you have a preference between bar charts and line graphs (or something else)?

ahawkes:
Thanks Tom - the stacked bar charts are way better for capacity - thanks for dealing with that.

alexdewar:
This is much clearer. Good work!

alexdewar (approved):

PR #297, opened 2024-05-09
Fix for models with no existing capacity

### LINKED ISSUE #286: [BUG] Error when running model with no existing capacity profile
## Describe the bug

I have a very simple model set up with only one sector with one technology. I seem to be getting the following error message:
'RuntimeError: Subsector with TRA_DMD_PCAR_PET_01 has no output commodities' 
This error seems to be appearing when the existing capacity file has only 0 values specified for all years of the simulation or 0 in the first year and an existing capacity specified from a later year onwards. If I specify a capacity in the first year or a retirement profile, I don't get any errors. Additionally I don't see the error If I set an arbitrarily small number in the first year. 
Also, I don't seem to be getting this error when I have a model set up with multiple sectors and technologies.

I am not quite sure what is causing the 'commodities' variable to be empty and have a length = 0 which leads to the error.

The full error message is: 

Traceback (most recent call last):
  File "/Users/sharwaridixit/anaconda3/envs/muse_env/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/sharwaridixit/anaconda3/envs/muse_env/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/sharwaridixit/.vscode/extensions/ms-python.debugpy-2024.0.0-darwin-arm64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/__main__.py", line 39, in <module>
    cli.main()
  File "/Users/sharwaridixit/.vscode/extensions/ms-python.debugpy-2024.0.0-darwin-arm64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 430, in main
    run()
  File "/Users/sharwaridixit/.vscode/extensions/ms-python.debugpy-2024.0.0-darwin-arm64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 284, in run_file
    runpy.run_path(target, run_name="__main__")
  File "/Users/sharwaridixit/.vscode/extensions/ms-python.debugpy-2024.0.0-darwin-arm64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 321, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/Users/sharwaridixit/.vscode/extensions/ms-python.debugpy-2024.0.0-darwin-arm64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 135, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/Users/sharwaridixit/.vscode/extensions/ms-python.debugpy-2024.0.0-darwin-arm64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 124, in _run_code
    exec(code, run_globals)
  File "/Users/sharwaridixit/Documents/Transport/Transport_trial_model_cars_e10/run_muse.py", line 13, in <module>
    model = MCA.factory(settings)
  File "/Users/sharwaridixit/MUSE_OS/src/muse/mca.py", line 74, in factory
    sectors.append(SECTORS_REGISTERED[kind](sector, settings))
  File "/Users/sharwaridixit/MUSE_OS/src/muse/sectors/sector.py", line 53, in factory
    subsectors = [
  File "/Users/sharwaridixit/MUSE_OS/src/muse/sectors/sector.py", line 54, in <listcomp>
    Subsector.factory(
  File "/Users/sharwaridixit/MUSE_OS/src/muse/sectors/subsector.py", line 224, in factory
    raise RuntimeError(msg)
RuntimeError: Subsector with TRA_DMD_PCAR_PET_01 has no output commodities

## To Reproduce

I have attached the folder with the input files; the existing capacity file currently has a 0 capacity for all the years therefore running the model should show the error.
[Transport_trial_model_cars_e10.zip](https://github.com/EnergySystemsModellingLab/MUSE_OS/files/15176419/Transport_trial_model_cars_e10.zip)

## Context

- Operative system (eg. Windows 10): MacOS
- MUSE version (eg. 1.0.1): Version 1.1.0
- Installation method (eg. pipx, pip, development mode): Virtual environment-based installation
- Python version (you can get this running `python --version`): 3.9.19

tsmbland:
# Description

this change.*

Fixes #286 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -30,7 +30,8 @@ def create_standard_agent(
         existing = capacity.interp(year=year, method=interpolation) > 0
         existing = existing.any([u for u in existing.dims if u != "asset"])
         years = [capacity.year.min().values, capacity.year.max().values]
-        capacity = xr.zeros_like(capacity.sel(asset=existing.values, year=years))
+        # capacity = xr.zeros_like(capacity.sel(asset=existing.values, year=years))
+        capacity = xr.zeros_like(capacity.sel(year=years))
     assets = xr.Dataset(dict(capacity=capacity))
     kwargs = _standardize_inputs(**kwargs)
 
@@ -118,9 +119,7 @@ def create_newcapa_agent(
 
     assets = xr.Dataset()
     if retrofit_present:
-        assets["capacity"] = xr.zeros_like(
-            capacity.sel(asset=existing.values, year=years)
-        )
+        assets["capacity"] = xr.zeros_like(capacity.sel(year=years))
     else:
         technologies = kwargs["technologies"]
         assets["capacity"] = _shared_capacity(
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -27,8 +27,10 @@ def create_standard_agent(
             technologies, capacity, region, share, year, interpolation=interpolation
         )
     else:
+        existing = capacity.interp(year=year, method=interpolation) > 0
+        existing = existing.any([u for u in existing.dims if u != "asset"])
         years = [capacity.year.min().values, capacity.year.max().values]
-        capacity = xr.zeros_like(capacity.sel(year=years))
+        capacity = xr.zeros_like(capacity.sel(asset=existing.values, year=years))
     assets = xr.Dataset(dict(capacity=capacity))
     kwargs = _standardize_inputs(**kwargs)
 
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -27,10 +27,7 @@ def create_standard_agent(
             technologies, capacity, region, share, year, interpolation=interpolation
         )
     else:
-        existing = capacity.interp(year=year, method=interpolation) > 0
-        existing = existing.any([u for u in existing.dims if u != "asset"])
         years = [capacity.year.min().values, capacity.year.max().values]
-        # capacity = xr.zeros_like(capacity.sel(asset=existing.values, year=years))
         capacity = xr.zeros_like(capacity.sel(year=years))
     assets = xr.Dataset(dict(capacity=capacity))
     kwargs = _standardize_inputs(**kwargs)
@@ -113,8 +110,6 @@ def create_newcapa_agent(
     if "region" in capacity.dims:
         capacity = capacity.sel(region=region)
 
-    existing = capacity.interp(year=year, method=interpolation) > 0
-    assert set(existing.dims) == {"asset"}
     years = [capacity.year.min().values, capacity.year.max().values]
 
     assets = xr.Dataset()
--- a/src/muse/utilities.py
+++ b/src/muse/utilities.py
@@ -255,9 +255,7 @@ def clean_assets(assets: xr.Dataset, years: Union[int, Sequence[int]]):
         x = set(assets.year[assets.year >= years].values)
         x.add(years)
         years = sorted(x)
-    result = assets.reindex(year=years, method="backfill").fillna(0)
-    not_asset = [u for u in result.dims if u != "asset"]
-    return result.sel(asset=result.capacity.any(not_asset))
+    return assets.reindex(year=years, method="backfill").fillna(0)
 
 
 def filter_input(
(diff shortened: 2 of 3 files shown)


tsmbland:
@dalonsoa I've been struggling a bit with this, so maybe you can help (see #286). In the `create_newcapa_agent` function it looks like the 'newcapa' agents are being limited to technologies that already exist, which doesn't seem right to me (unless I'm missunderstanding something). I've changed this, and changed the `clean_assets` function to allow empty assets, and this allows Sharwari's model to run. I'm not sure if the results make sense, but all the regression tests are now failing (not surprising since I've changed the model), so this probably isn't appropriate. What do you make of this?

I've made another PR (#298) which doesn't change the model and just changes the way that empty xarrays are dealt with to suppress the error. This also allows Sharwari's model to run, but in this case you have no capacity throughout the whole simulation as agents can't invest in technologies that don't already exist, which doesn't seem right either.

Any ideas? Am I barking up the wrong tree?

dalonsoa:
I think we have hit an edge case here - combined with a problem with the inputs.

First the second point. The simulation include retrofit agents which, by definition, only manage technologies with existing capacity. As there is no initial capacity, there's no point in having retrofit agents. This is fixed by doing the following:
- removing that line from the `Agents.csv` file
- setting the last column of the `Technodata.csv` file to `Agent1` (the only one we have).
- the section `sectors.road_vehicles.interactions` will need to be removed, as there are no new and retro agents to interact.
- the `demand_share` parameter in the toml file, section `[sectors.road_vehicles.subsectors.retro_and_new]` needs to be defined as `demand_share = "standard_demand"` (it defaults to `new_and_retro` if the parameter is not present, which requires retro agents). For some reason, though not surprisingly, this is totally undocumented!

These changes will let new capacity agents manage any existing capacity.

But this is not really the problem, just moves it to the new capacity agent. I don't think MUSE has ever been run in a situation where there's absolutely no initial capacity of any kind in any technology. It is assumed, and have checks for that, that here's always a non-null starting point.

To fix this, the only change needed in the code is in the calculation of the share of initial capacity between agents, ensuring that the returned array is not an empty one when there's no capacity:

https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/b6f17a4c171723b0fe682baeebcf6726d2686698/src/muse/agents/factories.py#L360

should change to:

```python
  if not any(techs):
      return (capacity * shares).copy()
  return (capacity * shares).sel(asset=techs.values).copy()
```

With the changes above, @sharwaridixit simulation runs beginning to end with non-null capacity being installed in all years. If they make sense or not that's a different story.

tsmbland:
@sharwaridixit Does this make sense? I've made Diego's suggested changes to the source code in #299. To try this out you will need to do a [source-code installation](https://muse-os.readthedocs.io/en/latest/installation/developers.html), and checkout the `no_capacity3` branch (`git checkout no_capacity3`), and also modify your model as outlined above. Let me know if this works and if the results make any sense.

sharwaridixit:
Thank you @dalonsoa @tsmbland for looking into the issue!
I applied the suggested changes to my model and ran it under the no_capacity3 branch. I don't seem to be running into any errors now and I checked the MCACapacity results which now seem to show capacity being built for the technology. The capacity for each year seems to match the demand outline in preset (I have attached an image of the results):
Screenshot 2024-05-10 at 11 42 31
However, I am a little unsure about why there are multiple entries of the capacity for the same year and am trying to look into this further; I was wondering if you had any idea about this?

tsmbland:
> However, I am a little unsure about why there are multiple entries of the capacity for the same year and am trying to look into this further; I was wondering if you had any idea about this?

As far as I'm aware this is normal for new versions of the code. I think at some point a change was made to have multiple rows in the output representing different assets. If you were using an older version of the code before then you may not have seen this, but if you sum all the rows for each year then this should be equivalent to how it was before. @dalonsoa do you know any more about this?

dalonsoa:
@tsmbland is correct. Each row represent a different asset, and it was changed fairly recently, but I'm not sure why it was done this way. I guess that they might differ in the installation year.

PR #303, opened 2024-05-14
.vscode/launch.json: Change types to debugpy

alexdewar:
It seems that using a type of `python` is deprecated and we should all be using `debugpy` instead. I'm not sure if VS Code's error message means that the debugging configs are currently being run with the old Python extension, but either way we should fix it.


dalonsoa (approved):
I haven't got any complaint about this, but if this works... so be it.

PR #304, opened 2024-05-14
Fix incorrect use of LCOE and ill-defined example

### LINKED ISSUE #246: [BUG] LOCE calculation is incorrect when more than one time slice is present
## Describe the bug
A clear and concise description of what the bug is, including error messages.

Email from Sara G: "to calculate the lcoe, all the costs are calculated and then allocated to the timeslices. Installation-related capital, fixed, and variable costs are therefore distributed over the timeslices. The costs are summed up and then divided over the energy produced by timeslice. When there is zero production in a timeslice, to avoid zero division in the algorithm, a very small epsilon amount is used. This can cause the big number in the output."

This is an error because a zero output in a time slice is a common situation. E.g. solar power at night.

The "big number" result should only happen when total output over the whole year approaches zero.

## To Reproduce
Steps to reproduce the behavior. Attache any input file that might be required.

Add a technology with zero output in a time slice. There needs to be more than one time slice.

## Expected behavior

The correct approach here is (I think - please feel free to pursue workable alterantives):

1) calculate variable cost (fuel, variable O&M, CO2, etc) in each time slice
2) calculate non-variable parts of cost (installation and fixed costs) and distribute across time slices in proportion to output. Note that if a time slice has zero output, none of these costs will be distributed to that time slice.
3) divide resulting value for each time slice by the output in that time slice (except where output is zero, to avoid a divide by zero error - if it's zero then set LCOE to zero for that time slice)

In the above we would also need to check what happens for cases where output is very very small but not zero - could be unstable.

## Context

- Operative system Windows 10 (probably all)
- MUSE version 1.1.0
- Installation method (eg. pipx, pip, development mode): All
- Python version (you can get this running `python --version`): n/a

dalonsoa:
# Description

This PR fixes a couple of problems:

1. The LCOE now returns a 0 for those timeslices where the utilization factor is zero - which would result in an infinity LCOE as a result. Fixes #246 .
2. The `default_timeslice` model was ill-defined, with wrong values for the `MinimumServiceFactor`, which must be between 0 and 1, inclusive.
3. Fixes a weird bug that was reveled when changing the above two. It's unclear why things were not failing before, but I suspect that another, unrelated error was causing the associated tests to pass. The real problem is that the tests that were supposed to check this are really more integration tests that actual unit tests, so a lot of things were going on. 

## Type of change

## Key checklist

## Further checks


ahawkes:
Thanks this looks good to me.
Is there another issue required to fix the input data on the time slice problem?

alexdewar:
One regression test is failing for all the runners... I'm guessing we're getting a different result now because the calculation has changed @dalonsoa?

src/muse/investments.py
```diff
@@ -346,7 +346,7 @@ def scipy_match_demand(
         )
         if not res.success:
             getLogger(__name__).critical(res.message)
-            print(f"in sector containing {df_technologies.technology[0]}")
+            print(f"in sector containing {df_technologies.technology.unique()}")
```
alexdewar, line 349: We don't really want to be using `print` for error messages:

```suggestion
```

Alternatively you could include it in the logging message above if you think it's useful information.

src/muse/objectives.py
```diff
@@ -240,21 +259,14 @@ def capacity_to_service_demand(
     **kwargs,
 ) -> xr.DataArray:
     """Minimum capacity required to fulfill the demand."""
-    from muse.timeslices import represent_hours
```
alexdewar, line 262: pydocstyle will complain about this empty line:

```suggestion
```

src/muse/objectives.py
```diff
@@ -664,6 +678,7 @@ def lifetime_levelized_cost_of_energy(
         + fixed_and_variable_costs
     ) / (denominator.sel(commodity=products).sum("commodity") * rates).sum("year")
 
+    results = results.where(np.isfinite(results)).fillna(0.0)
     return results
```
alexdewar, line 682:
```suggestion
    return results.where(np.isfinite(results)).fillna(0.0)
```

alexdewar (approved):
A few minor suggestions, but otherwise all good!

src/muse/readers/csv.py
```diff
@@ -918,11 +918,7 @@ def check_utilization_not_all_zero(data, filename):
     else:
```
alexdewar, line 918: You could get rid of the `else` and remove some indentation here

src/muse/readers/csv.py
```diff
@@ -918,11 +918,7 @@ def check_utilization_not_all_zero(data, filename):
...
         utilization_sum = data.groupby(["technology", "region", "year"]).sum()
 
-        # Add small value to 0 utilization factors to avoid numerical problems
-        if utilization_sum.utilization_factor.any() == 0:
-            data.loc[data.utilization_factor == 0, "utilization_factor"] = (
-                data.loc[data.utilization_factor == 0, "utilization_factor"] + 0.01
-            )
+        if (utilization_sum.utilization_factor == 0).any():
```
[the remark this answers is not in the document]
dalonsoa: It's a column in a pandas dataframe. And yes, this conditional was broken before. I cannot understand how it was working.

dalonsoa:
> Thanks this looks good to me. Is there another issue required to fix the input data on the time slice problem?

As far as I can tell, that's all. The results seems to make sense and once the minimum service issue is solved and meaningful values are provided for it, the calculation completes even with some utilization factors equal to zero, which I think was the next main barrier.

I would suggest we merge this, create a release and you try to use it to check that it produces sound results.

src/muse/readers/csv.py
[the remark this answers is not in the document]
alexdewar: Yeah, that was my thought :laughing:

ahawkes:
OK thanks. Maybe Tom could look at the impact on the tutorials, to try to see if it looks like the result make sense?

Note, however, that the overall approach is still lacking despite this fix. This is because LCOE (and presumably other objectives) is calculated based on maximum utilisation. But when solved, actual utilisation could be a lot lower (e.g. the technology might be used only in 1 time slice), thus increasing LCOE substantially, potentially (likely!) making the technology choice wrong ex-post of the dispatch.

Rgds, Adam

ahawkes:
I think it would require substantial changes, and indeed there is no perfect solution that I am aware of. In MUSE 1.0, suggest we leave this as-is for now, as a known issue. In MUSE 2.0 I am looking at it, and I think we can get close to a solution that considers actual post-dispatch LCOE at time of investment, but it will never be perfect (much like the real world, which in some senses is what MUSE was developed to do!) and will also probably present some new issues.

dalonsoa:
@tsmbland it might be simpler if you review this and we merge it, and then you update your branch with develop. Splitting it into smaller bits is not going to help was you still need to bring all changes to your branch, anyway.

tsmbland (approved):
Looks good to me. I think there's only one small change I'll have to make to one of the tutorials

PR #305, opened 2024-05-14
Enable ruff's pydocstyle checks

### LINKED ISSUE #292: `ruff`: Enable `pydocstyle` checks
These checks make sure docstrings are correctly formatted. We probably want "google" style as it's closest to what's already there.

I tried running `ruff` with these checks enabled and most of the complaints were about missing docstrings for functions and modules. We could just disable these checks for now and deal with them later.

alexdewar:
Enables `ruff`'s pydocstyle checks and fix the errors. I've left warnings about missing
docstrings disabled, because adding all of them would be a rather large undertaking.

For missing argument descriptions I've done my best, but there was a bit of guesswork
involved in places.

Fixes #292.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,8 +102,12 @@ ignore_errors = true
 
 [tool.ruff.lint]
 select = [
+    "D",  # pydocstyle
     "E",  # pycodestyle
     "F",  # Pyflakes
     "I"  # isort
 ]
-ignore = []
+ignore = [
+    "D1"  # missing docstring checks
+]
+pydocstyle.convention = "google"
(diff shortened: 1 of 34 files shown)


alexdewar:
....aaaaand the tests are failing. I'll look into it

dalonsoa (approved):
I've made a couple of suggestions of things that looked odd, but otherwise, looks good. 

An important clarification - which might confuse other people as well - the demand is of commodities. Technologies require or produce commodities when working, but the demand is of those commodities, not the technologies themselves. They are a means to an end.

(Big tech companies often try to misled us, offering new technologies as if we need them, while what we need is the commodities those technologies enable - often none extra compared with the previous generation of the same technology.)

dalonsoa:
And tests are failing, but you already knew that 😄

PR #307, opened 2024-05-16
VS Code: --no-cov argument no longer needed for pytest

alexdewar:
The --no-cov argument is not only no longer necessary when invoking `pytest` from VS Code, but in this case, as we don't have `pytest-cov` installed, it means that the command will fail and VS Code will be unable to run the tests from the tests panel.

On another note, do we want to add `pytest-cov` to the dependencies? And on yet another note, do we want to install the codecov GitHub app so we can see coverage changes for PRs?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,8 +17,6 @@
     "editor.formatOnSave": true,
     "editor.rulers": [88],
 
-    // --no-cov needed so that breakpoints work: https://github.com/microsoft/vscode-python/issues/693
-    "python.testing.pytestArgs": ["--no-cov"],
     "python.testing.unittestEnabled": false,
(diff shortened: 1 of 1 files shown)


dalonsoa (approved):
I don't think we need `pytest-cov` as a dependency.

Coverage reports [should be generated and uploaded to Codecov](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/1451b269b249c4a683309dcaac5f7195ebc7404c/.github/workflows/ci.yml#L47) automatically, in principle when running the CI.

Happy to install the GitHub app.

alexdewar:
I think the codecov app can do that too, as well as annotating PRs. I'll have a look at it

dalonsoa:
Done!

Issue #295, opened 2024-05-07
Tutorial 8.7 (timeslice constraints) - example figure doesn't abide by maximum timeslice constraints


tsmbland:
I'm not yet sure what the source of the error is, but the figure at the bottom doesn't appear to abide by the maximum service constraints described in the text (unless I'm misunderstanding something).

tsmbland:
Needs #335

tsmbland:
This now appears to be fixed. #264 still required to clarify the notebook

Issue #264, opened 2024-04-17
MinimumServiceFactor tutorial needs clarification (8.7.1)


tsmbland:
The MinimumServiceFactor tutorial is confusing as the meaning of this parameter isn't properly explained (i.e. what does a specific number mean?). 

The explanation should be extended to explain what a specific number means. If there is a fuller explanation of this parameter elsewhere in the documentation, this should also be linked.

PR #312, opened 2024-05-17
Update tutorial notebook text

### LINKED ISSUE #311: Update notebook text to match updated models
As of #279 the tutorial models are up to date with the latest version of the default model, and the steps necessary to generate these models are fully documented as python scripts. We now need to update the text of the tutorial notebooks accordingly. Things to pay attention to:
- Make sure that all the steps necessary to generate the models are documented in the notebooks, so that users have all the information they need to replicate the results
- Make sure that the starting point for the tutorial is clear (i.e. building off the default model vs an earlier example)
- Some of the tables will need updating
- Some paths referencing the model files may need updating
- Some file names have changed (e.g. technodata.csv -> Technodata.csv)

tsmbland:
# Description

This PR makes a number of changes to the tutorial notebooks and underlying models. The tutorials are by no means finished, this is more just a step in the right direction, but this PR is getting quite large now (sorry), so I think it's worth reviewing and merging.

The main purpose of this PR was to update the text in the notebooks to make sure they have all the information needed to replicate the results. I have also made a few changes (simplifications) to the underlying models which will hopefully make it easier for users to follow the tutorials.

At some point we will want someone with domain knowledge to go through all of these and make sure they make sense, but I think this can wait until the remaining issues are completed.

Main things still to do:
- #337
- #295
- #264
- #340 

---

Summary of changes:

General changes:
- fixing links
- fixing minor typos (e.g. technodata.csv -> Technodata.csv, projections.csv -> Projections.csv)
- be more specific about the starting point for each tutorial
- changes to tables so they actually show the right numbers

Section 4:
- shrinking plots so they can be viewed side-by-side

Section 8.1:
- simplify model to be more in line with `default`
- add note about `excluded_commodities` in `settings.toml`
- show data for all sectors

Section 8.2:
- switch order of models to have the simpler scenario first
- the population wasn't being split properly between both agents, this is now fixed
- show data for all sectors

Section 8.4: 
- simplify the consumption profile
- a few changes to growth/capacity parameters
- add note about `forecast` parameter in `settings.toml`

Section 8.5:
- rename model back to what it originally was (fixes some links which were broken)
- a few changes to growth/capacity parameters

Section 8.6:
- plot sectors side-by-side

Section 8.7:
- clarify the starting point as the `default_timeslices` model. This is a change compared to before, where `default` was the starting point, and should make it much easier for the user
- set `MinimumServiceFactor` to zero by default (this will be unecessary once #334 is done)

---

It's probably easiest to review by taking a glance at the documentation site for this branch [(here)](https://muse-os.readthedocs.io/en/documentation/), but bear in mind that tutorials 2 and 7 won't make much sense until the remaining issues are fixed

Fixes #311 

## Type of change

## Key checklist

## Further checks


tsmbland:
Might be a good one for @HarmonicReflux to review (i.e. work through the new notebooks and make sure you're able to replicate the results)

tsmbland:
Pesky regression tests failing on macos - not sure what to do about this...

alexdewar:
The tests appear not to be running... Does anyone have any idea why?

dalonsoa:
No idea. They were running yesterday, I think.

dalonsoa:
I've just closed and re-open the PR. That has triggered the workflows again.

dalonsoa:
And now, everything fails... :(

tsmbland:
Numpy 2.0, yay!

dalonsoa:
Here we go...

alexdewar:
FYI There's a `ruff` rule to upgrade to numpy 2.0 (`NPY201`): https://numpy.org/devdocs/numpy_2_0_migration_guide.html

docs/user-guide/modify-timing-data.ipynb
```diff
@@ -42,36 +44,35 @@
...
+    "We do the same for the ```Residential2050Consumption.csv```, but set the demand in 2050 to be triple that of 2020 in every timeslice. See [here](https://github.com/SGIModel/MUSE_OS/blob/main/docs/tutorial-code/4-modify-timing-data/1-modify-timeslices/technodata/preset/Residential2050Consumption.csv) for the full file.\n",
     " \n",
     "<!-- As we have modified the timeslicing arrangements there will be a change in the underlying demand for heating. This may require more electricity to service the demand for heat. Therefore, we relax the constraints for growth in ```windturbine``` in region `R2` in the ```technodata/power/Technodata.csv``` file, as well as the constraints for ```heatpump```, again in region `R2`. For the full power sector ```Technodata.csv``` file click [here INSERT LINK HERE](github-power-technodata), and for the full residential sector ```Technodata.csv``` file click [here INSERT LINK HERE](github-residential-technodata). -->"
```
alexdewar, line 77: I know you haven't modified this line, but as it's commented out, maybe we should drop it:

```suggestion
```

docs/user-guide/modify-timing-data.ipynb
```diff
@@ -169,35 +172,40 @@
...
     "The `time_framework` details each benchmark year in which we run the simulation. The ```foresight``` variable details how much foresight an agent has when making investments. \n",
     "\n",
-    "As we have modified the timeslicing arrangements there will be a change in the underlying demand for heating. This may require more electricity to service this demand. Therefore, we relax the constraints for growth in the power sector for all technologies and constraints in the ```technodata/power/Technodata.csv```, as shown below:\n",
+    "We also have to modity the `forecast` value for each sector. For example, for the power sector the `sectors.power.subsectors.retro_and_new` section should now look as follows:\n",
+    "\n",
+    "    [sectors.power.subsectors.retro_and_new]\n",
+    "    agents = '{path}/technodata/Agents.csv'\n",
+    "    existing_capacity = '{path}/technodata/power/ExistingCapacity.csv'\n",
+    "    lpsolver = \"scipy\"\n",
+    "    forecast = 2\n",
```
alexdewar, line 181: You could do this as a GitHub-style code block, i.e. starting with \`\`\`toml, so you get proper syntax highlighting.

alexdewar:
I've only just noticed that we're not pinning dependency versions... yikes.

How about for now we just change the numpy requirement to be `numpy<2` so that we can finish this PR and #328? We can always fix the numpy thing later.

alexdewar:
PS I tried the `ruff` rule and it seems it may not be the panacea I was hoping. It only identified one problem:

```patch
diff --git a/src/muse/quantities.py b/src/muse/quantities.py
index 604286d5..1fa4cad8 100644
--- a/src/muse/quantities.py
+++ b/src/muse/quantities.py
@@ -574,7 +574,7 @@ def supply_cost(
         else:
             data = data.groupby("region").sum(asset_dim)
 
-    total = data.production.where(np.abs(data.production) > 1e-15, np.infty).sum(
+    total = data.production.where(np.abs(data.production) > 1e-15, np.inf).sum(
         "timeslice"
     )
     return data.prices / total
```

tsmbland:
Thanks @alexdewar for the good points! I've made the changes you suggested.

Yeah, I agree that I should have broken this up a bit more into multiple PRs - my bad. Hopefully future changes will be smaller.

I've pinned numpy to <2.0 (and opened #345), but I'm still having issues with some of the tests (unrelated to numpy 2.0). It's just that some of the results on macos are ever-so-slightly different to the expected results, but I don't know why this is or how to fix it (maybe we just need to change the tolerance?)

I've seen issues like this before and usually you can fix it by re-running the tests, but this seems particularly stubborn...

alexdewar:
> Yeah, I agree that I should have broken this up a bit more into multiple PRs - my bad. Hopefully future changes will be smaller.
> 

Nw!

> I've pinned numpy to <2.0 (and opened #345), but I'm still having issues with some of the tests (unrelated to numpy 2.0). It's just that some of the results on macos are ever-so-slightly different to the expected results, but I don't know why this is or how to fix it (maybe we just need to change the tolerance?)
> 
> I've seen issues like this before and usually you can fix it by re-running the tests, but this seems particularly stubborn...

I think increasing the tolerance might be the way we have to go. Unfortunately there are subtle differences in the way different processor manufacturers design their floating-point units and so you often get slightly different results on Intel/AMD/ARM. (Come to think of it, I've got an AMD processor here, so maybe I should try running the regression tests too...)

tsmbland:
Fixed the failing tests by increasing rtol to `1e-4` (which I think is still small enough to be meaningful)

alexdewar (approved):
LGTM!

PR #327, opened 2024-06-05
Revert "Enforce minimum service factor limits"

alexdewar:
Reverts EnergySystemsModellingLab/MUSE_OS#325

It seems that the tests are now failing on `develop` and I didn't catch this before merging (see #306). Sorry!


alexdewar:
So I made the changes you suggested on the last PR, pushed them and clicked auto-merge, then it merged before actually running the tests, many of which actually failed :upside_down_face:

PR #328, opened 2024-06-05
Enforce minimum service factor limits (take 2)

### LINKED ISSUE #320: Invalid values for minimum service factor in `test_minimum_service_factor`
As established in #304, `MinimumServiceFactor` should be between 0 and 1. However, the `test_minimum_service_factor` test uses invalid values >1, and still passes

Ideally we shouldn't be in a position where the model can be run with invalid values, however at the very least this test should be modified to use valid values.

It's also worth making sure this test is checking the correct thing. It seems to be comparing the absolute value of supply against the `MinimumServiceFactor`, however my understanding is that `MinimumServiceFactor` relates to relative supply compared to the theoretical maximum (hence why it needs to be between 0 and 1). Or maybe I'm misunderstanding something...

alexdewar:
Let's try this again. It turns out the fix for my previous mistake was really trivial (5fb3ff5824ae034bd11955006001a5dde321d897). Sorry about all the noise!

Fixes #320.

### DIFF
--- a/src/muse/readers/csv.py
+++ b/src/muse/readers/csv.py
@@ -98,8 +98,9 @@ def read_technodictionary(filename: Union[str, Path]) -> xr.Dataset:
     data.columns.name = "technodata"
     data.index.name = "technology"
     data = data.drop(["process_name", "region_name", "time"], axis=1)
-
     data = data.apply(to_numeric, axis=0)
+
+    check_utilization_not_all_zero(data, filename)
 
     result = xr.Dataset.from_dataframe(data.sort_index())
     if "fuel" in result.variables:
--- a/src/muse/readers/csv.py
+++ b/src/muse/readers/csv.py
@@ -130,6 +130,9 @@ def read_technodictionary(filename: Union[str, Path]) -> xr.Dataset:
 
     if "year" in result.dims and len(result.year) == 1:
         result = result.isel(year=0, drop=True)
+
+    check_minimum_service_factors_in_range(data, filename)
+
     return result
 
 
@@ -145,7 +148,7 @@ def read_technodata_timeslices(filename: Union[str, Path]) -> xr.Dataset:
     data = csv[csv.technology != "Unit"]
 
     data = data.apply(to_numeric)
-    data = check_utilization_not_all_zero(data, filename)
+    check_utilization_not_all_zero(data, filename)
 
     ts = pd.MultiIndex.from_frame(
         data.drop(
@@ -934,4 +937,12 @@ def check_utilization_not_all_zero(data, filename):
             f"""A technology can not have a utilization factor of 0 for every
                 timeslice. Please check file {filename}."""
         )
-    return data
+
+
+def check_minimum_service_factors_in_range(data, filename):
+    min_service_factor = data["minimum_service_factor"]
+    if not np.all((0 <= min_service_factor) & (min_service_factor <= 1)):
+        raise ValueError(
+            f"""Minimum service factor values must all be between 0 and 1 inclusive.
+             Please check file {filename}."""
+        )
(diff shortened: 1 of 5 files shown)


src/muse/readers/csv.py
```diff
@@ -129,7 +133,7 @@ def read_technodata_timeslices(filename: Union[Text, Path]) -> xr.Dataset:
     data = csv[csv.technology != "Unit"]
 
     data = data.apply(lambda x: pd.to_numeric(x, errors="ignore"))
-    data = check_utilization_not_all_zero(data, filename)
+    check_utilization_not_all_zero(data, filename)
```
dalonsoa, line 136: in principle, the `technodata_timeslices` file also has an `utilization_factor` column that will need validating. See https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/enforce_minimum_service_factor_limits/src/muse/data/example/default_timeslice/technodata/power/TechnodataTimeslices.csv

tests/test_minimum_service.py
```diff
@@ -1,36 +1,37 @@
...
+    permutations((np.linspace(0, 1, 6), [0] * 6)),
 )
-def test_minimum_service_factor(tmpdir, minimum_service_factor, process_name):
+def test_minimum_service_factor(tmpdir, minimum_service_factors):
     import pandas as pd
     from muse import examples
     from muse.mca import MCA
 
     sector = "power"
+    processes = ("gasCCGT", "windturbine")
```
dalonsoa, line 34: Concrete values of the service factor aside, this does not check the same scenarios as before. Before there were 4 tests being launched, and now they are only 2. Not that the previous 4 cases were necessary, but just flagging this.

dalonsoa (changes requested):
The changes look ok, but I've a couple of comments/concerns, and a request.

tsmbland (changes requested):
Looks good, although I think we also need to have a similar check for `MinimumServiceFactor` in the `read_technodata_timeslices` function. While we're at it, it would also be good to have a similar check for `UtilizationFactor` (i.e. it needs to be between 0 and 1). And we could also check that `UtilizationFactor` is always >= `MinimumServiceFactor`.

Also, it might be nice to have a some tests with invalid inputs to make sure these checks fail when they should.

alexdewar:
All good points @tsmbland!

tests/test_minimum_service.py
alexdewar: There should still be 4... `permutations()` returns the inputs in 2 different orders

alexdewar:
Ok, I think I've made all the changes that were suggested and I've added a bunch more tests. I'm glad I had a more thorough look at this because it turns out my original fix didn't actually work :worried:

dalonsoa (approved):
Looks good! Way more comprehensive validation than before.

dalonsoa:
But tests are failing...

alexdewar:
Sorry :disappointed:. I pushed this at the end of the day and figured it was all fine (the tests I was working on were passing!). I'll take another look. Hopefully it's just something trivial!

alexdewar:
Converting to draft until I fix the test failures (see also #362)

tests/test_readers.py
```diff
@@ -410,3 +412,150 @@ def test_read_trade_technodata(tmp_path):
...
+        mock.assert_called_once_with(df, "file.csv")
+    min_service_factor_mock.assert_not_called()
+    utilization_below_min_mock.assert_not_called()
+
+
+@patch("muse.readers.csv._check_utilization_in_range")
+@patch("muse.readers.csv._check_utilization_not_all_zero")
+@patch("muse.readers.csv._check_utilization_not_below_minimum")
+@patch("muse.readers.csv._check_minimum_service_factors_in_range")
+def test_check_utilization_and_minimum_service_factors_fail_missing_utilization(*mocks):
```
tsmbland, line 547: I don't think you need the mocks here. Otherwise, everything looks good!

tsmbland (approved):

Issue #332, opened 2024-06-10
Identify necessary outputs from the input data db


tsmbland:
We need to think about what data we will need to retrieve from the database and what format it will take. This will probably be xarrays which are identical to objects already existing in MUSE (see readers/csv.py). We should identify what these objects are, including a list of dimensions, variables, and data types.

It's probably better to look into this sooner rather than later as it might inform some design choices for the database.

To do:
- Update docstrings for the functions in `csv.py` to give more info about the xarray objects returned (dimensions, variables, data types)
- Add tests in `test_readers.py` to check the outputs of these functions
- Identify a suitable subsection of the [schema](https://lucid.app/lucidchart/0fa3cd6c-aa69-4397-b432-024be7acb22e/edit?invitationId=inv_c2989e36-7e9d-4865-abf5-d258fd73b274&page=0_0#) to begin implementing a minimal database, and the expected outputs from this database. I've suggested the commodities section (blue tables in the schema) as an option, but open to other ideas

tsmbland:
Closed by #346

PR #346, opened 2024-06-18
Expand reader tests

cc-a:
# Description

As prepatory work for #332 this PR expands the tests for the reader functions in `readers/csv.py` to more strictly codify the expected output. So far I've only looked at `tests/test_readers.py::test_read_trade_technodata` and wanted to get initial feedback before working on others. Particularly interested on anything else about the structure of the dataset that could be checked.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_readers.py
+++ b/tests/test_readers.py
@@ -607,3 +607,53 @@ def test_read_csv_agent_parameters(default_model):
             "share": "agent_share_2",
         },
     ]
+
+
+def test_read_initial_market(default_model):
+    from muse.readers.csv import read_initial_market
+    from muse.readers.toml import read_settings
+
+    settings = read_settings(default_model / "settings.toml")
+    path = default_model / "input" / "Projections.csv"
+    data = read_initial_market(path, timeslices=settings.timeslices)
+
+    assert isinstance(data, xr.Dataset)
+    assert set(data.dims) == {"region", "year", "commodity", "timeslice"}
+    assert dict(data.dtypes) == dict(
+        prices=np.float64,
+        exports=np.float64,
+        imports=np.float64,
+        static_trade=np.float64,
+    )
+    assert list(data.coords["region"].values) == ["R1"]
+    assert list(data.coords["year"].values) == list(range(2010, 2105, 5))
+    assert list(data.coords["commodity"].values) == [
+        "electricity",
+        "gas",
+        "heat",
+        "CO2f",
+        "wind",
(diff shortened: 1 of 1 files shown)


tsmbland:
This looks good! I don't think there's anything else that needs to be checked

I've realised I don't actually know what the `TradeTechnodata.csv` file is for as it isn't included in the `default` example which is what I've been working off, so the data for this specific example probably isn't included in the schema for the new files. You'll want to base future functions off the `default` model (i.e. `copy_model("default", tmp_path)`) where possible.

In theory #312 is done, but I'll leave it up until after the standup tomorrow in case anyone else wants to review it. It would also be easy just to pin the numpy version in this PR or a separate one.

tsmbland (approved):

cc-a:
This PR attempts attempts a comprehensive pass at the reader functions in an attempt to more rigourously define the data model used within Muse. Where possible reader functions are tested against the default example model using a single example data file. Other models are used where necessary. Some functions it was not possible/necessary to test for the reason given.

Reader functions tested against default model (and the data file used):
- read_technodictionary - "technodata/residential/Technodata.csv"
- read_io_technodata - "technodata/residential/CommOut.csv"
- read_initial_assets - "technodata/residential/ExistingCapacity.csv"
- read_global_commodities - "input/GlobalCommodities.csv"
- read_csv_agent_parameters - "technodata/Agents.csv"
- read_initial_market -  "input/Projections.csv"
- test_read_attribute_table - "input/Projections.csv"

Reader functions tested against other models:
- read_technodata_timeslices - default_timeslice - "technodata/power/TechnodataTimeslices.csv"

Reader funtions not tested:
- read_initial_capacity - called from multiple examples but is always passed an array and never reads data from a file
- read_technologies - wraps other reader functions, does not read data directly
- read_csv_timeslices - not called by any example model
- read_timeslice_shares - not called by any example model
- read_macro_drivers - not called by any example

Potentially these last may need to be reviewed as to whether they are still relevant or could be removed.

cc-a:
Thanks @tsmbland. You make a good point. I think for your current purposes recording the tightest possible type makes sense particularly as it's hard to know how interchangeable e.g. ints and floats are in the code. But we should think carefully about how we tightly we specify types when writing tests for the planned changes.

tsmbland (approved):
Looks good! I think we can merge what you've got here and create a separate branch for the new inputs

PR #349, opened 2024-06-19
Update examples and tutorials to not use retrofit agents

### LINKED ISSUE #340: Update examples/tutorials to NOT using retrofit agents
## Is your feature request related to a problem? Please describe

All examples and tutorials use the pair of agents newcapa/retorfit, but that is no longer necessary and MUSE works totally fine with just new capacity agents managing all the assets. However, this is not documented and no examples show how to use MUSE this way. 

Remove retrofit agents from all examples but 1 or 2.

Keep using retrofit agents. Retrofit agents might be useful but can be confusing, specially for new users, so better start with simpler scenarios. 

## Additional context

N/A

### LINKED ISSUE #350: Change default value for `demand_share` to "standard_demand"
The `demand_share` parameter defines how the demand is distributed among agents. The current default is `new_and_retro`, which should be used when there are new and retrofit agents. When only new agents are being used, `standard_demand` should be used. Since we're removing retrofit agents from the default models, we should update the default `demand_share` to `standard_demand`.

tsmbland:
# Description

Sorry for the massive PR - this ended up being a bigger job than I expected. Most of the changes, however, are to results files.

The point of this PR is to change most of the example and tutorial models to use agents with a single 'new' share, rather than a pair of 'new' and 'retrofit' shares. This is a more likely scenario for basic users, and makes the tutorials simpler.

This PR also includes the changes made in #386 to increase the `maximum_iterations` parameter.

Main changes:
- Change the example models to use `standard_demand` instead of `new_and_retro`, remove the `new_to_retro` interaction, remove the `Agent2` retrofit agent share, and allocate all existing capacity to the `Agent1` (new) agentshare
- These changes have been propagated to the tutorial models, which are all based on the `default` and default_timeslice` models
- Add a `default_retro` model to the examples, which is equivalent to the old `default` model
- The only example model I haven't changed is the `trade` model, as I couldn't get it to work without the retrofit agents.
- Change the default value of `demand_share` to `standard_demand`
- Modify the `add_agent` parameter in the `wizard` module to work without retrofit agents
- Update tests to work with the new models
- Changes to the tutorial notebooks in line with the changes to the models. You can see the new version of the tutorials [here](https://muse-os.readthedocs.io/en/documentation/).

It would probably be a good idea to have a tutorial using retrofit agents in the advanced guide, but I think that should be a separate issue.
	

Fixes #340 
Fixes #350 

## Type of change

## Key checklist

## Further checks


tsmbland:
I've just updated the inputs for some of the models so far (haven't regenerated the results), but does this look reasonable? @dalonsoa

dalonsoa (approved):
I think these changes look sensible, so hopefully they should result in feasible models. Have you tried to run them? Do they result in any error?

tsmbland:
Yeah they all run and the results look similar to before, so I think it's all good

tsmbland:
I have no idea why the regressions tests are failing. @dalonsoa @alexdewar Can you try running tutorial 4.2 locally in this branch and let me know if you get any diffs?

dalonsoa:
This is failing also locally for me. I think it makes sense. The time framework in the settings is `[2020, 2022, 2024, 2026, 2028, 2030, 2032, 2034, 2036, 2038, 2040]` but in the `Results` folder that is use as the "correct" values there are also 2025, 2035, 2045 and 2050, which I guess should not be there. As the tests compares the "correct" output with the new output, it fails. You will need to remove the files that should not be there anymore. 

image

tsmbland:
Makes sense! Thanks

alexdewar (approved):
I've only given this a quick once-over, but it seems sensible.

dalonsoa (approved):
Now that it passes, all looks good!

tsmbland:
I've just realised that changing the default demand share to "standard_demand" could be a breaking change for models where demand share isn't explicitly stated in the settings file. Should I change it back to "new_and_retro"? @dalonsoa @ahawkes

alexdewar:
I guess we have to change it back if it's a breaking change :disappointed:.

We could issue a warning if people don't specify the demand share though, saying something like "it's defaulting to `new_and_retro` for backwards compatibility reasons but you probably don't want to use this"

tsmbland:
Broader point - but how essential is it that we don't make breaking changes, since we know everyone that's using the software and can warn them of these sorts of things? (ignoring the technicalities of semantic versioning)

ahawkes:
I think we should go ahead with this one and just handle the breaks that happen. Not sure I fully understand the extent of potential problems, but I do really want to get away from new_and_retro

PR #354, opened 2024-06-20
Enable customisation of constraints

### LINKED ISSUE #336: New constraints are ignored
When working on #323 , I've found another issue...

## Problem

In theory, it is possible to add new constrains simply y adding an appropriately coded and decorated function and then indicating in the `toml` file to use it:

```
@register_constraint
def new_constraint(...):
    ...
```

And then, in the relevant subsector of the `toml` file, add:

```
[sectors.residential.subsectors.retro_and_new]
...
constraints = [
    ...,
    "new_constraint"
]
```

However, **these settings are ignored unless the agent is of type `Agent`**; `retrofit` and `newcapa` agents, the only ones we use as far as I can tell, are `InvestingAgents` and they DO NOT use this setting but use instead a [hardcoded list of constraints](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/7c8036caf12509af46e998ea361c34cd0b454e27/src/muse/constraints.py#L229) - all of them. 

## Solution

To enable customisation of the `InvestmingAgents` constraints, the `agents_factory` needs to be enabled to use that and pass the appropriate parameter to [the function call in the subsectors factory](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/7c8036caf12509af46e998ea361c34cd0b454e27/src/muse/sectors/subsector.py#L184). These constraints settings might need to be propagated down to the appropriate level to ensure they are used by the agents.

dalonsoa:
# Description

Enables the customisation of constraints, which should be there but it was not. So far, constraints to be used were hardcoded and always the same, being the settings in relation to constraints passed in the TOML file ignored. This PR fixes that issue. 

The settings file where the constraints sections include `# Optional, defaults to the constraints below` have been updated, if needed, to include the really default constraints that were being used so far. 

Fixes #336 

## Type of change

## Key checklist

## Further checks


tsmbland (approved):
I've looked through the code and I _think_ this will work, although the `constraints` list gets passed around to so many different classes and factories it took a while to figure out what's going on!

A test might be nice to confirm that it works, but I'll leave that up to you

alexdewar (approved):
LGTM too, though I haven't traced the constraints setting through the code...

Do the results look different now?

dalonsoa:
The results are not affected because we are using the same constraints - the only thing this does is letting the user actually select which constrains are to be used, which was not possible before. 

I've checked this works by using the debugger and seeing that, indeed, now changing the input constraints results in different constraints being called in the relevant places (and because I've working in #355 which needs this, and it works), but I agree a test should be in place. 

I'll add a test before merging.

dalonsoa:
I've added a test to check that the input constraints settings are indeed passed to the agents factory, which was the only barrier, really.

PR #355, opened 2024-06-20
Adds demand-limitting capacity constraint

### LINKED ISSUE #323: Combined capacity should exactly match the maximum demand of all timeslices
No more, no less - the combined capacity should be equal to the maximum demand required in any timeslice. Two options to implement this:

- Implement an inequality constrain that enforces that the capacity must not exceed the maximum demand.
- Implement an equality constrain the strictly enforces this. 

Both will be explored in terms of ease of implementation and performance.

dalonsoa:
# Description

Adds a new (default) constraint that limits the combined capacity being installed to just matching the demand, no more. While such a condition should be reached automatically in general by minimising the costs, it should not harm to impose it via an upper limit constraint as there might be several local minima and have the solver falling in one of them.

The constraint is setup such that utilisation factors are taken into account, so if there is only a technology with an utilisation factor of 0.1 in a timeslice, then the installed capacity can be up to 10 times the peak demand, such that the production can indeed satisfy the demand.

Only tutorial `docs\tutorial-code\2-add-agent\2-multiple-objective` has been affected by this new constraint. As seen in the modified output files, the result is a drastic redistribution of the investment by each agent, now both looking very similar, with a reduction of the combined total capacity. I'm not entirely sure if this is reasonable or not. I guess that if we are now constraining the installed capacity, the influence on the different decision objectives is less relevant, as the search space is much smaller... @ahawkes , what do you think?

image

 I've removed the specific list of constraints for the residential sector in the TOML file for this tutorial so it uses the default settings, like the other sectors.

Fixes #323 

## Type of change

## Key checklist

## Further checks


tsmbland:
I haven't looked at the code yet, but the results look much more reasonable now. Agent 2 was massively overinvesting before because it uses EAC which is negative, so could 'minimise costs' by overinvesting. This is no longer an option. I guess EAC and LCOE are similar enough in how they rank the technologies that Agent 1 and 2 look very similar when this constraint is imposed. In this case it probably makes more sense to use a different objective for the tutorial so there's more of a difference between the two agents, but that's a future job for me.

src/muse/constraints.py
```diff
@@ -493,6 +494,64 @@ def max_production(
...
+        demand_constraint.b.max("timeslice")
+        if "timeslice" in demand_constraint.b.dims
+        else demand_constraint.b
+    )
+
+    # Now we need to find the maximum capacity constraint (as the capacity here is
+    # negative), and switch the sign to make it positive. In practice, we want to
+    # capture the utilisation factor of the technologies.
+    capacity = (
+        -capacity_constraint.capacity.max("timeslice")
```
tsmbland, line 537: Why the `max` operation? I guess capacity will be the same in all timeslices so this is just to reduce the data over the timeslice dimension? Unless I'm misunderstanding something

src/muse/constraints.py
```diff
@@ -493,6 +494,64 @@ def max_production(
...
+
+    # We need to find the most demanding timeslice and use as the demand constraint
+    b = (
+        demand_constraint.b.max("timeslice")
+        if "timeslice" in demand_constraint.b.dims
+        else demand_constraint.b
+    )
+
+    # Now we need to find the maximum capacity constraint (as the capacity here is
+    # negative), and switch the sign to make it positive. In practice, we want to
```
tsmbland, line 534: Why is capacity negative here? I realise there's a minus sign in the final line of `max_production`, but I don't really understand why that's done...

src/muse/data/example/default/settings.toml
```diff
@@ -60,7 +60,8 @@ constraints = [
     "max_capacity_expansion",
     "demand",
     "search_space",
-    "minimum_service"
+    "minimum_service",
+    "demand_limitting_capacity"
```
tsmbland, line 64: Just a note that this change will also change the settings files for the tutorial models once `generate_models.py` is run (although this won't change the results), so may as well do that now

src/muse/data/example/default_timeslice/settings.toml
```diff
@@ -62,7 +62,8 @@ constraints = [
     "max_capacity_expansion",
     "demand",
     "search_space",
-    "minimum_service"
+    "minimum_service",
+    "demand_limitting_capacity"
```
tsmbland, line 66: As above

src/muse/constraints.py
dalonsoa: The capacity is not really "capacity", but the capacity-related component of the constraint, which is a multiplying factor to the actual capacity being installed. The capacity itself is, naturally, the same for all timeslices but this factor also includes the utilisation factor, which can be timeslice dependent. By taking the "max" we make sure that the timeslice with the minimum utilisation factor is covered.

src/muse/constraints.py
dalonsoa: Again, not capacity, but a multiplier to to the capacity. And it is negative because we are using as the starting point the maximum production to avoid repeat Bing the same calculation. In there, the capacity related factor is negative given that the production related factor is positive, but here we need it positive, so the equation "Cf x Capacity <= Demand" has a solution with a positive capacity.

src/muse/data/example/default/settings.toml
dalonsoa: Didn't I change the tutorials files as well? My fault if I didn't...

src/muse/data/example/default_timeslice/settings.toml
dalonsoa: Also as above 😊

tsmbland:
I think this mostly makes sense but I'm still confused about a few things. For example, what happens if you have two timeslices, one with demand=1 and UF=1, and the other with demand=0.1 and UF=0.1. It looks like capacity would then be allowed to reach the point where it can meet demand=1 with a UF of 0.1 (which is much more than it actually needs)? Is that right?

dalonsoa:
That's a very good point. In that case, the capacity constraint will be much higher than needed. I'll review the code to account for that circumstance.

dalonsoa:
@tsmbland , I've updated this so it implements a more complex approach to find the right timeslice to use in the constraint. This still fixes the tutorial 2, as explained above, but also affects tutorial 7. Other than that, everything else remains the same.

tsmbland (approved):
Looks good! Not sure what's going on with the regression tests on macos...

alexdewar:
The floating-point errors seem on the larger side... I'm not sure either, but @jamesturner246 did his PhD on this!

@jamesturner246 We're seeing weirdly large differences between the Ubuntu/Windows runners and the (ARM) macOS ones... Any ideas? https://github.com/EnergySystemsModellingLab/MUSE_OS/actions/runs/9760527594/job/26942272480?pr=355

alexdewar (approved):
LGTM (except for the mysterious failing MacOS runners....)

One small thing: it should be "limiting" rather than "limit**t**ing" (yes, English spelling makes no sense!)

tsmbland:
@dalonsoa I'll try merging the latest changes from the dev branch and see if that fixes the tests

jamesturner246:
> @jamesturner246 We're seeing weirdly large differences between the Ubuntu/Windows runners and the (ARM) macOS ones... Any ideas? https://github.com/EnergySystemsModellingLab/MUSE_OS/actions/runs/9760527594/job/26942272480?pr=355

Not sure on the surface.. Can't see the runner log any more. 

Some leads to begin with:

* Are these ARM64 M1 runners, or AMD64? 
* Is the macos compiler passing in a sneaky `-ffast-math`?
* Clang/stdlib version?
* Does a Windows/Ubuntu clang runner have comparable error?
* Any other arch dependent features/dependencies?

jamesturner246:
Furthermore, if any `long double`s are used, ARM chips use 64 bits, whereas x86_64 historically **might** use legacy extended precision 80 bit x87 registers. Check your FP types.

PR #356, opened 2024-06-20
Enable pyupgrade in ruff and fix the issues that pop up

### LINKED ISSUE #293: `ruff`: Enable `pyupgrade` checks
We should wait for @dalonsoa to finish #289 before this, as we should be able to bump the project's minimum Python version at this point.

dalonsoa:
# Description

`pyupgrade` is now enabled in `ruff`, using python 3.9 as the target since we want to keep the compatibility with it. As usual with these things, a lot of files are changed, but otherwise you should only worried about `pyproject.toml`.

Fixes #293 

## Type of change

## Key checklist

## Further checks


docs/advanced-guide/extending-muse.ipynb
```diff
@@ -89,7 +89,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from typing import List, Optional, Text\n",
+    "from typing import Optional\n",
```
alexdewar, line 92: I guess we have to wait for a newer version of Python until we can write `Optional`s as `my_type | None`!

src/muse/constraints.py
```diff
@@ -397,9 +390,9 @@ def demand(
     search_space: xr.DataArray,
     market: xr.Dataset,
     technologies: xr.Dataset,
-    year: Optional[int] = None,
+    year: int | None = None,
```
alexdewar, line 393: I think writing union types this way is a Python 3.10 thing, alas: https://docs.python.org/3/library/typing.html#typing.Union

alexdewar: It's a bit weird that the Python 3.9 runners don't seem to be complaining... Maybe these bits aren't covered by tests?

alexdewar (changes requested):
I've skimmed this and it mostly looks totally fine (and the newer syntax is much nicer to look at!).

I am a little concerned that there are some places where you're writing union types as `type_a | type_b`, which is only available in Python 3.10. Oddly, none of our workflows have caught this (maybe because we don't have explicit tests for the type hints?), but I'm a bit concerned things will break at runtime for Python 3.9 users.

src/muse/constraints.py
dalonsoa: These sort of things are checked by `mypy` which we are not running in MUSE because it largely lacks of consistent typing information. Maybe at runtime these things do not matter at all?

What annoys me is that I set as target python version 3.9, so if this is a 3.10 feature, it should not have been changed :(

alexdewar: Ugh, that is annoying. I guess it's a `ruff` bug? :disappointed: 

Strangely, there are plenty of other places where it's left `Optional[int]`s alone.

I suppose one option would be to do this first upgrade with `pyupgrade` itself, which presumably doesn't have this problem? I guess the `ruff` hook might then change these bits again, though. Perhaps if it's being really stubborn we should just use the `pyupgrade` pre-commit hook.

dalonsoa: I've just found the trick in [here](https://stackoverflow.com/a/76660598). Esentially, you can use `|` with no problem from python 3.7 as long as you do `from __future__ import annotations` which we do in all the files that `ruff` changed. 

In summary, these changes are safe to go.

dalonsoa: In the file where that was not changed it was because we did not have that import statement.

alexdewar: Ohhh I see. Well that's a relief

alexdewar (approved):
LGTM!

tsmbland (approved):

PR #365, opened 2024-06-21
Tests fail when running the full testing suite

### LINKED ISSUE #261: Tests fail when running the full testing suite [BUG]
## Describe the bug

The developer documentation instructs to use `python -m pytest` to run the tests. However, I've found that the tests fail when using this command (more details below). I note that this command isn't actually run as part of the CI, which first runs the tests excluding the notebook tests and the regression tests, before running the regression tests separately. Splitting the tests up in this way also works fine locally. Nevertheless, the tests should ideally pass when running the full suite. Otherwise, the developer instructions should change to clarify which tests should be run locally and which shouldn't.

Details:
- Two regression tests fail when running the full test suite with the command above. Here's the error:
```
FAILED tests/test_fullsim_regression.py::test_tutorial_regression[tutorial_path10] - KeyError: "['timeslice', 'asset', 'year', 'commodity'] not in index"
FAILED tests/test_fullsim_regression.py::test_tutorial_regression[tutorial_path11] - KeyError: "['timeslice', 'asset', 'year', 'commodity'] not in index"
```
- However, these test pass when running `pytest tests/test_fullsim_regression.py`
- `pytest -m "not regression and not notebook"` passes
- I've also noticed that various files for simulation results in `docs/advanced-guide` are changed during the tests. Why are these files changing? Should these changes be committed? The notebooks `extending-muse.ipynb` and `further-extending-muse.ipynb` appears to be the culprits.
- Why aren't notebook tests being run as part of CI?
- I _think_ the models that are failing are `0-new-decision-metric` and `3-add-region`

## To Reproduce
I ran the following:
```
python -m venv .venv
.venv/scripts/Activate.ps1
pip install -e .[dev,doc]
python -m pytest
```

## Expected behavior
Ideally the tests should pass when running the full suite. Otherwise, the developer instructions should change to clarify which tests should be run locally

## Context

- Operative system (eg. Windows 10): I've tried on Windows 10
- MUSE version (eg. 1.0.1): dev branch
- Installation method (eg. pipx, pip, development mode): dev mode
- Python version (you can get this running `python --version`): 3.9.18

HarmonicReflux:
# Description

This pull request refactors the way `muse --model -default` is executed to ensure that all tests pass.

Specifically, the Results/MCACapacity.csv file is now placed within the docs folder, which is where the test expects to find MCACapacity.csv. The necessary file is created during the test run by the notebook itself and subsequently picked up by it. This change resolves the error that causes pytest to fail, and as a result, all tests now pass by default.

Fixes #261 

## Type of change

## Key checklist

## Further checks


docs/installation/developers.rst
```diff
@@ -14,20 +16,29 @@ Once you have ``git`` in your system, clone the MUSE repository with:
...
 
     cd MUSE_OS
-    # 1- Create virtual environment
-    # 2- Activate virtual environment
-    # 3- Finally, install MUSE in editable mode with:
-    python -m pip install -e .[dev,doc]
+    # 1- Create a virtual environment
+    # 2- Activate that virtual environment
+    # 3- Install MUSE in editable mode with: python -m pip install -e .[dev,doc]
+    # 4 - Invoke `muse --model default`
```
dalonsoa, line 33: So this is the key thing to enable all tests to pass when the full suite is run at once, right?

dalonsoa (approved):
I've re-run the failed test and now things work as planned. It is just a flaky test.

The changes look good and the explanations are clearer now, so all good.  

For future PR remember to 1) provide a description of what the PRs is about, 2) indicate what issue it is closing and 3) make sure the branch is up to date.

docs/installation/developers.rst
tsmbland: I don't think so, as this will generate a results folder in the current working directory, whereas we specifically need it in the `docs` folder

tests/test_docs_notebooks.py
```diff
@@ -7,6 +7,9 @@
 
 def available_notebooks() -> list[Path]:
     """Locate the available notebooks in the docs."""
+    if not Path("Results/MCACapacity.csv").exists():
+        return []
```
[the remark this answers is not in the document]
dalonsoa: Sorry, I missed this file. Yes, we do not want this. All tests should run.

dalonsoa (changes requested):
All tests should run, and not skipping them if a file is missing. If it is missing, we need to figure out why it is missing and put it in there.

docs/installation/developers.rst
HarmonicReflux: Yes. For the tests to pass without the failure one has to run `muse --model default`.
Independently, there are is still an "expected fail", some tests that got "skipped" and four "warnings", though I opened separate issues for them.

dalonsoa:
I think the main issue here is:
1. to figure out what notebook needs the Results folder in the root folder
2. then take the steps such that the data it looks for is within the docs directory - like any other of the notebooks - and put the data in the right place. 
3. And yes, committing the data as well, as it is done with the other notebooks that need data, like those of the tutorials.

What makes no sense is to have a notebook in the docs requiring data in the root directory. 

There's a [discussion going on](https://github.com/EnergySystemsModellingLab/MUSE_OS/discussions/351) on how to avoid committing data to the repository, but that's a longer term discussion and we do not have an answer for it, yet.

dalonsoa:
And remember to update the branch and, ideally, add a PR description - not essential, but serve as reference for the future.

alexdewar:
Did you want me to review this @HarmonicReflux?

HarmonicReflux:
> Did you want me to review this @HarmonicReflux?
Diego already put forth some requested changes, so I am good for now, thanks.

HarmonicReflux:
> Did you want me to review this @HarmonicReflux?

Yes, please take a look, so we are all on the same page. 
Thks.

docs/installation/developers.rst
alexdewar: > In the docs workflow [a link is made between both](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/d331f46c13a70889ca093492ea441848aec14a79/.github/workflows/documentation.yml#L29), but it is true I'm not entirely sure that's a solution in this case.

Yeah, that was a hack on my part to avoid having to generate results twice.

I think it's a good idea to tell users that they need to run this command in order for tests to pass, but I think we should put that in the "Running Tests" section, not here.

alexdewar (changes requested):
As the others have said, we don't want to disable tests when the files aren't there. While that will make the tests pass, that'll just mean that developers who haven't run the default model first won't run these tests at all, so they won't know if their code is broken until they open a PR.

I think the right solution is to just put an instruction about this in the documentation for now. We could potentially do something cleverer, like automatically running the default model in `conftest.py` if the files are missing, but we shouldn't do that on this PR. I'd maybe open an issue called "default model must be run for tests to pass" and we can have a think about it. (We don't want to go back to committing the contents of `Results/` to the repo, because we've only just got rid of them!)

As @dalonsoa has said, you want a PR description so reviewers know what has been changed and why. In this case, it would have been good to have a heads-up that you'd disabled some of the tests! If @tsmbland hadn't clocked it then it could have broken things down the line.

HarmonicReflux:
Appreciated comments of the reviewers.

alexdewar (changes requested):
I still think the text about having to run `muse --model default` should be in the "Running Tests" section. See comment

HarmonicReflux:
@alexdewar please take a look at this reworked pull request.

tsmbland:
I'm still a bit unsure about this, because we specifically need a results folder in the `docs` folder, so running `muse --model default` in the root folder won't be enough. The link that Alex created in the docs workflow will only exist if the user runs the documentation build locally (right?) which we're not asking people to do before running the tests, so I'm not sure how this fixes the problem. Am I missing something?

HarmonicReflux:
Running a model will create a "Results" folder in the directory MUSE is installed. The results folder itself then contains subfolder that of which one contains the file the one testing the notebooks needs to have access in order to run successfully. To my knowledge, building the documentation is a separate process unrelated to the tests themselves.

tsmbland:
I get that, but the notebook in question (`running-muse-example.ipynb`) points to a Results folder contained in the `docs` folder, not the root folder

HarmonicReflux:
@dalonsoa and @tsmbland,
please review this pull-request. 

@alexdewar is out of office until later this week.

tsmbland (approved):

dalonsoa (approved):
The latest modification is very neat! Let's hope it works in all instances we are interested.

alexdewar (approved):
LGTM!

PR #368, opened 2024-06-24
Min production considered in supply

### LINKED ISSUE #335: Fix the `supply` function in `quantities.py` to impose minimum constraint
I've just found why the discrepancy.

1. The sector output, taking the values right out of the market, gives the right values as market information is calculated based on the constrains (so figure on the right is correct).
2. The MCA information, surprisingly and I don't know why, is obtained by recalculating everything. In [this line](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/7c8036caf12509af46e998ea361c34cd0b454e27/src/muse/quantities.py#L91), it limits the supply so it is not higher than the maximum production, but it does not impose any minimum constrain but to fulfil the demand, and therefore the minimum service constrained is not honoured.

In summary, the calculation for the `MCAMetric_Supply.csv` is wrong - and I would presume that any other calculation that uses [this supply function](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/7c8036caf12509af46e998ea361c34cd0b454e27/src/muse/quantities.py#L14) will also be wrong, as no minimum constrains are imposed but to fulfill the demand.

_Originally posted by @dalonsoa in https://github.com/EnergySystemsModellingLab/MUSE_OS/discussions/321#discussioncomment-9749169_

dalonsoa:
# Description

Adds a minimum service constraint to the supply calculation, so it is consistent with the constraints used in the investment. This change only affects to the outputs of a couple of examples and tutorials - and only the `MCAMetric`-related outputs -, but none of the others, still producing the same outputs.

Fixes #335 , identified and further discussed in #321 

Why there are two measures of supply instead of using the production calculated by the solver is another question and a matter for a different PR. After this PR is merged, both figures in #321 become the same, and equal to the one on the right. 

## Type of change

## Key checklist

## Further checks


src/muse/quantities.py
```diff
@@ -47,6 +47,7 @@ def supply(
         production_method = maximum_production
 
     maxprod = production_method(technologies, capacity)
+    minprod = minimum_production(technologies, capacity)
```
alexdewar, line 50: Would it make sense to rename `production_method` to `maximum_production`? It might be clearer.

src/muse/quantities.py
```diff
@@ -660,3 +663,65 @@ def group_assets(x: xr.DataArray) -> xr.DataArray:
...
+    The dimensions above are only indicative. The function should work with many
+    different input values, e.g. with capacities expanded over time-slices :math:`t` or
+    agents :math:`i`.
+
+    Arguments:
+        capacity: Capacity of each technology of interest. In practice, the capacity can
+            refer to asset capacity, the max capacity, or the capacity-in-use.
+        technologies: xr.Dataset describing the features of the technologies of
+            interests.  It should contain `fixed_outputs` and `minimum_service_factor`.
+            It's shape is matched to `capacity` using `muse.utilities.broadcast_techs`.
```
alexdewar, line 689:
```suggestion
            Its shape is matched to `capacity` using `muse.utilities.broadcast_techs`.
```

alexdewar (approved):
Seems sensible!

dalonsoa:
Not fixing the problem.

tsmbland:
@dalonsoa I still think we should do this. There's no reason not to include the MSF in this function, and as discussed in #423, this function _is_ being used to calculate supply even when using the scipy solver (except, as discussed in #474, when using `dispatch_production = "costed"`)

tsmbland (approved):

PR #374, opened 2024-06-26
Remove Level parameter from Technodata files

tsmbland:
# Description

`Level` is included as a parameter in the Technodata files, but this parameter isn't actually used anywhere in the model (it's just dropped when the file is read).

It will be clearer if we remove this parameter.

(The `Level` parameter in the `CommIn` and `CommOut` files _is_ used, so this stays)

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Maybe it was used by some obscure, never finally implemented neither documented functionality...

PR #377, opened 2024-06-27
Remove "Level" parameter for outputs and update documentation

### LINKED ISSUE #263: Meaning of "fixed" vs "flexible" level is unclear
The documentation mentions that the "Level" parameter in the technodata and commodities files can be "fixed" or "flexible", but the meaning of these two options isn't adequately explained.

Sections to look at:
9.3.3 Techno-data (`docs/inputs/technodata.rst`)
9.3.5 Commodities (`docs/inputs/commodities_io.rst`)

Update: #374 has changed the technodata files (and documentation) to no longer include the Level parameter as it wasn't being used

tsmbland:
# Description

I have removed the 'Level' column from the CommOut files, as this parameter only applies to input commodities and not output commodities (all output commodities must be 'fixed' type).

I have also updated the documentation to give some more details about what this parameter is actually doing.

Fixes #263 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.9
+    rev: v0.4.10
     hooks:
       - id: ruff-format
         types_or: [python, pyi, jupyter]


docs/inputs/commodities_io.rst
dalonsoa, line 1: Ah... Now it makes sense :)

src/muse/commodities.py
```diff
@@ -66,17 +66,9 @@ def just_tech(x):
...
             return x.any(dims)
 
-        products = list(
-            just_tech(technologies[x] > 0)
-            for x in {"fixed_outputs", "flexible_outputs"}
-            if x in technologies.data_vars
-        )
-        if len(products) == 0:
+        if "fixed_outputs" not in technologies.data_vars:
             raise ValueError("Missing output array in technologies")
```
dalonsoa, line 70: Maybe indicate what array is the one missing:
```suggestion
            raise ValueError("Missing 'fixed_outputs' array in technologies")
```
Or something along these lines?

src/muse/readers/csv.py
```diff
@@ -181,6 +181,9 @@ def read_io_technodata(filename: Union[str, Path]) -> xr.Dataset:
     from muse.readers import camel_to_snake
 
     csv = pd.read_csv(filename, float_precision="high", low_memory=False)
+    if "Level" not in csv.columns:
+        # Applies to outputs files, where level is not specified (must be "fixed")
+        csv["Level"] = "fixed"
```
dalonsoa, line 186: In principle, this function is applied to both input and output, so how do you know this particular bit of code will apply only to the output file?

dalonsoa (changes requested):
I've a couple of suggestions, but otherwise it looks good - specially the explanation in the docs.

src/muse/readers/csv.py
tsmbland: If the user follows on from one of the example models, then the "Level" column will only be in the inputs file and not the outputs file.

If the outputs file _does_ have this column, then this won't run, but as long as all the values are "fixed" then it doesn't make a difference (and any "flexible" values will get flagged later on anyway).

If the inputs file _doesn't_ have this column, I think it's also fine as "fixed" is reasonable as a default value.

I should probably clarify this in the comment

dalonsoa: Yeah, I think it is the last point the one that was unclear to me. If that can be explained, it would be great. In general, I think we need to make explicit what any default values in the simulation are.

dalonsoa (approved):

PR #380, opened 2024-06-28
Duckdb with sqlalchemy

cc-a:
# Description

This PR builds on #379 prototypes a partial adoption of SQLAlchemy in combination with Duckdb. This attempts to get the best of both worlds by using SQLAlchemy to define the database schema whilst still leveraging the capabilities of Duckdb to work with CSV files and easily spit out numpy arrays.

The main point in using SQLAlchemy here is to cut down on the amount of SQL that needs to be written and to do schema definition via the Python declarative interface. By having a generic CSV reader function you can then cut down to two simple lines of SQL. The advantage of this may be more apparent in a more mature implementation that will require more complex SQL in the duckdb only approach.

One very nice side benefit of having the schema in SQLAlchemy would be on the input creation side. At the moment the only real approach is to hack a bunch of csv files and deal with errors as you try to read them in. Instead you could use the SQLAlchemy classes to populate a database then dump it into csv files which could be quite powerful in the case of large or complex input datasets.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,9 @@ dependencies = [
     "pypubsub",
     "tomlkit",
     "duckdb",
-    "fsspec"
+    "fsspec",
+    "sqlalchemy",
+    "duckdb-engine"
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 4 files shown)


tsmbland:
Funnily enough I actually find the SQL table schemas more readable than the SQLAlchemy typed classes, but I get your point about more complex tables.

RE input creation - that reminds me: One thing that we will want to do is the reverse process of populating the database using data from the xarrays, and then creating csv files from the database. This is so we can essentially convert input files from the old format to the new format (i.e. old format files -> xarrays -> database -> new format files). We should have a go at doing this before scaling up the database as this may influence our approach.

PR #381, opened 2024-07-01
Remove any hardcoded reference to "month", "day" and "hour"

### LINKED ISSUE #363: Timeslice level names are no longer configurable [BUG]
## Describe the bug

Previously you could choose the name of the timeslice levels using the `level_names` argument in `settings.toml`. However, since #289, the default level names ("month", "day" and "hour") are hardcoded in many different places in the code, but we still accept the `level_names` argument in `settings.toml`. If anyone tries to change this, the model will fail.

I think before it would also accept fewer or more timeslice levels (at least according to the documentation, although I haven't tried this), but obviously now it's restricted to three levels.

## To Reproduce

Try changing `level_names` for any of the example models

## Expected behavior

We should either change the code to stop hardcoding level names, or remove the `level_names` option from `settings.toml` so the defaults ("month", "day" and "hour") are mandated.

I'd probably go with the latter, as it doesn't seem like there would be much use for changing the level names, but maybe I'm wrong?

## Context

- MUSE version (eg. 1.0.1):
- Installation method (eg. pipx, pip, development mode):

### LINKED ISSUE #382: `test_minimum_service_factor` is not actually testing anything
In the final for-loop of this test it's filtering the `MCAMetric_Supply` dataframe by timeslice and then checking that the supply for each timeslice is greater than the minimum service factor.

The problem is, it's using integers to filter by timeslice, whereas the timeslice values are actually strings (e.g. `('all-year', 'all-week', 'night')`). Thus, the filtering produces an empty dataframe, so there's nothing to compare against the minimum service factor and the test just passes regardless. To see this, try changing `>= factor` to `>= np.inf` and the test will still pass!

The other problem is that we shouldn't be directly comparing supply against the minimum service factor, as the MSF is relative to capacity (so supply vs capacity * MSF would be the appropriate comparison)

tsmbland:
# Description

After #392, there are still a couple of places in the code that referenced "month", "day" and "hour" by name, which could cause problems if the level names were changed from the default (which is possible to in the settings file). This started off as a PR to remove all of these hardcoded references, but this ended up having a few other implications, so this PR now addresses a few things.

Specific changes:

`scipy_match_demand` in `investments.py`. This was easy to fix as it was just dropping the "month", "day" and "hour" columns needlessly, so I just removed this.

`sector_supply` and `sector_consumption` in `mca.py` which prepare a couple of output files. I found a workaround to do a similar thing without referring to "month", "day" and "hour" by name, and in the process slightly modified the output files (for the better, I think).

This second change then allowed `test_minimum_service` to start actually testing something (see #382), but the test was failing because it was testing the wrong thing. So I fixed the test so it's now testing the right thing, although the test is still failing, presumably because the supply data that it's using is wrong (see #321, #335 and #368). So not sure what to do now. @dalonsoa If we're planning on fixing the broken supply output, are you happy for me to mark this test as xfail for now so we can merge this?

Fixes #363 
Fixes #382

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
The changes look totally sensible so I'm happy to move on with this one and marking the test as `xfail` if you're sure that the test is failing because it is not correctly setup and not because the changes to the functions you have made have broken the execution pipeline.

So we don't forget, also open an issue about it, assign it to me and add it to this iteration.

src/muse/outputs/mca.py
```diff
@@ -643,26 +634,19 @@ def sector_consumption(
...
-                ).reset_index()
-                b = b[b["consumption"] != 0]
-                data_sector.append(b)
-    if len(data_sector) > 0:
-        output = pd.concat([u for u in data_sector], sort=True)
+            if not a.empty:
+                data_sector.append(a[a["consumption"] != 0])
 
+    if len(data_sector) > 0:
+        output = pd.concat([u for u in data_sector], sort=True).reset_index()
```
alexdewar, line 646: Does this need to be a generator expression?

alexdewar (approved):
Seems fine to me.

tsmbland:
@dalonsoa I've marked the test as xfail. We already have #335 which should fix the test when it's sorted, but I'm not sure what we've doing about that one.

Issue #383, opened 2024-07-01
LCOE output doesn't work [BUG] 


tsmbland:
## Describe the bug

You get an error if you try to include the LCOE output:

`ValueError: cannot insert hour, already exists`

Raised by [this line](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/bd73930d60d46824ffcf7a47befa66b95ab9a207/src/muse/outputs/mca.py#L1123)

## To Reproduce

Include the following in the settings file for a model

```toml
[[outputs]]
quantity = "LCOE"
sink = "aggregate"
filename = "{cwd}/{default_output_dir}/MCA{Quantity}.csv"
```
and you will get the error above

## Expected behavior

This should work without error, and create a file containing the LCOE metric for the simulation. This worked when I tried it a few weeks ago investigating #316, so whatever the problem it's relatively new. I think it was very likely introduced by #289

## Context

Please, complete the following to better understand the system you are using to run MUSE.

- Operating system (eg. Windows 10):
- MUSE version (eg. 1.0.1): dev
- Installation method (eg. pipx, pip, development mode): dev
- Python version (you can get this running `python --version`): 3.12

tsmbland:
And it seems LCOE beats supply by being calculated in __three__ different places!

Issue #338, opened 2024-06-14
[BUG] Supply is much larger than annual preset demand 


sharwaridixit:
## Describe the bug

I have set up a model which describes the transport sector including subsectors for road vehicles, aviation and rail. After running the model, I found in the MCASupply results file that the sum of the supply for the end-use demand commodities for each simulation year, except the base year, seems to be much higher than the specified demand for that year. For example, in 2020 the supply for TC is 1321.6876 bvkm whereas the TC demand for 2020 is 440.5624 bvkm; this supply result is also 3 times larger than the demand. This is also the case for the other commodities as well. I am not sure as to why this is happening and why the difference is a factor of 3 ( I cannot seem to identify any input value in the model that would lead to this).

## Screenshots
The first image shows the supply results and the image below it shows the demand file.
<img width="743" alt="Screenshot 2024-06-14 at 15 16 17" src="https://github.com/EnergySystemsModellingLab/MUSE_OS/assets/147044537/38c8a96c-f771-413f-9732-346b00fb8e42">
<img width="1280" alt="Screenshot 2024-06-14 at 15 26 09" src="https://github.com/EnergySystemsModellingLab/MUSE_OS/assets/147044537/a31f7d0e-3f50-4485-b943-7d52c82c704f">

## To Reproduce

I have attached the zip file containing my model:
[Transporttrialmodel.zip](https://github.com/user-attachments/files/15839744/Transporttrialmodel.zip)

## Expected behavior
I expected the supply to be equal to the demand specified in the preset file.

## Context

Please, complete the following to better understand the system you are using to run MUSE.

- Operative system (eg. Windows 10): MacOS
- MUSE version (eg. 1.0.1): Version 1.1.0
- Installation method (eg. pipx, pip, development mode): Virtual environment-based installation
- Python version (you can get this running `python --version`): 3.9.19

tsmbland:
Hi Sharwari, I'm trying to run your model, but it's failing for me with an error (`ValueError: zero-size array to reduction operation fmax which has no identity`). I'm using python 3.9 and MUSE 1.1.0. Could you please let me know which versions of numpy and xarray you're using?

sharwaridixit:
Hi @tsmbland, 
I am using version 2022.3.0 for xarray and version 1.23.0 for numpy.

tsmbland:
[...]

PR #384, opened 2024-07-01
Enable consumption data in `demand_path` format

### LINKED ISSUE #364: No longer compatible with consumption data in the `demand_path` format [BUG]
## Describe the bug

I tried running the model from #338 with the current dev version of MUSE, and faced a number of issues.

The first is an issue with how timeslices are defined in `settings.toml`. MUSE expects three levels for timeslice, but only two levels were defined for each of the timeslices in the file (e.g. `winter.night` should be `winter.all-week.night` etc). I'm not sure why this error is permitted in old versions of MUSE and not the dev version. This is a good thing and not a problem with MUSE, I'm just flagging this first as it's relevant for tacking this issue.

Anyway, when I fix this issue, the model still fails to run with the following error:

`ValueError: These variables cannot be found in this dataset: ['day', 'timeslice', 'month', 'hour']`

which relates to `drop` in line 118 of `preset_sectors.py`.

This seems to relate to the format that the consumption data takes in the input files. All the models we've been working with use the `consumption_path` format which includes timeslice data, however this model uses the `demand_path` format which doesn't include any timeslice data (see https://muse-os.readthedocs.io/en/latest/inputs/toml.html#preset-sectors)

I guess this is why it's complaining about trying to drop the timeslice variables, as these variables don't exist in the input data, but I'd have to dig into it a bit further to find the exact problem

## To Reproduce

Try running the model from #338, fixing the error in `settings.toml` that I described above 

## Expected behavior

The model should run without error

## Context

- Python version (you can get this running `python --version`): 3.12

tsmbland:
# Description

This is a not very elegant way of getting Sharwari's model from #338 to work with the latest MUSE version. The problem was that she was specifying the demand data in a format that doesn't have timeslice information (the `demand_path` format, see [here](https://muse-os.readthedocs.io/en/latest/inputs/toml.html#preset-sectors)), which is a valid scenario and something that worked before, but something that we accidentally broke with (I think) #289.

Specifically, a couple of places in the code were complaining because it was trying to call `drop_vars(["timeslice", "month", "day", "hour"])` on xarrays that don't have timeslice data. Adding `errors="ignore"` means that it will try to drop these variables, and not raise a fuss if they already don't exist.

See also #381 which is also related to these `drop_vars` statements.

Ideally we'd have a test to make sure this doesn't break again, although I'm not sure the best way to do that without creating a new example model for the regressions tests, which seems a bit overkill. @dalonsoa - ideas?

Note also that Sharwari's model still fails due to #383 (although that one could be easily bypassed by removing the output)

Fixes #364 

## Type of change

## Key checklist

## Further checks


alexdewar (approved):
LGTM!

In terms of how we catch things like this in future, I agree that we probably don't want another example model, but it might still be useful to have @sharwaridixit's model be run as part of the regression tests, if it's using code paths that are otherwise untested. Would it be useful to have another set of models which are *just* used for those tests? Alternatively, I guess we could add a unit test or two.

dalonsoa:
I think that, at this point, we might want to recover @alexdewar idea of a helper function to drop these. Something like:

```python
def drop_timeslice(data):
    if "timeslice" not in data.dims:
        return data

    return data.drop_vars([c for c in data.timeslice.indexes])
```

This would also be independent on the name of the timeslice levels. Normally they are `month`, `day`, `hour`, but that is just a convention. One could equally use `season`, `week`, `hour`.

tsmbland:
Yes that's a good idea. I'll give that a try in #381

tsmbland:
Opened #392 to deal with timeslice dropping. Just need this very small change as well to fix the issue

PR #386, opened 2024-07-01
Increase maximum_iterations parameter to 100

### LINKED ISSUE #372: `maximum_iterations` value for the example/tutorial models?
I hadn't quite appreciated the importance of the `maximum_iterations` parameter before, and have just gone with 1 for all of the tutorials as this is what the `default` model was already using. However, after reading [this page of the documentation](https://muse-os.readthedocs.io/en/latest/inputs/projections.html#initial-market-projection), it seems there's quite a fundamental difference between a value of 1 and anything >1.

@ahawkes do you have a sense of what we should be using for `maximum_iterations` in the tutorial/example models?

It currently defaults to 3 if it isn't specified in `settings.toml`, but I'm not sure how that number was arrived at.

tsmbland:
# Description

The example models were previously using a range of values for the `maximum_iterations` parameter, ranging from 1 to 100. As explained in the issue (#372), this should be a high number, so I've changed all the models to use 100 and regenerated the results files.

I've also changed the default value from 3 to 100 (i.e. when not specified in the settings file)

I've based this PR off #349 as that one is already changing all the results files. There may be some text changes required in the notebooks (e.g. describing plots), but I'll address these sorts of things in #349 once this is merged

Fixes #372 

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Looks sensible to me. It's interesting to see how the output files change, adding more assets installed in the future years. I hope those changes still make sense!

alexdewar (approved):
LGTM!

PR #387, opened 2024-07-02
Use Even Bettter TOML

dalonsoa:
# Description

As Better TOML plugin has been deprecated - and by VS Code keeps complaining about it. 

The `prettier` pre-commit hook repo has also been archived recently which suggests there will be no more updates and it will remain in a `v4.0.0-alpha.8` state forever. Nothing to do now, but just pointing it in case it causes trouble in the future.

Fixes # N/A

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,7 +3,7 @@
     // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
     // List of extensions which should be recommended for users of this workspace.
     "recommendations": [
-        "bungcip.better-toml",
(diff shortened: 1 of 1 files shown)


alexdewar:
> The `prettier` pre-commit hook repo has also been archived recently which suggests there will be no more updates and it will remain in a `v4.0.0-alpha.8` state forever. Nothing to do now, but just pointing it in case it causes trouble in the future.

That's a bit annoying. There are other hooks for formatting YAML and JSON which we could use, but `prettier` also has a VS Code extension so they don't end up fighting...

PR #388, opened 2024-07-02
Correctly use the minimum service factor

### LINKED ISSUE #375: Minimum service factor is applied incorrectly
Currently, we seem to be multiplying utilisation factor by minimum service factor, which is not the correct calculation. (I guess this means that the minimum is treated as being a proportion of utilisation factor?)

See: https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/a2a1b4f3d8bf4166ee57056b569584234a454883/src/muse/constraints.py#L533

@dalonsoa I may have misremembered, but did you say that there were other places in the code where we're using this value incorrectly?

dalonsoa:
# Description

As ti was being used as a fraction of the utilisation factor rather as a fraction of 100%.

Fixes #375 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/constraints.py
+++ b/src/muse/constraints.py
@@ -530,7 +530,7 @@ def minimum_service(
         .drop_vars("technology")
     )
     capacity = convert_timeslice(
-        techs.fixed_outputs * techs.utilization_factor * techs.minimum_service_factor,
+        techs.fixed_outputs * techs.minimum_service_factor,
         market.timeslice,
         QuantityType.EXTENSIVE,
     )
--- a/docs/inputs/technodata.rst
+++ b/docs/inputs/technodata.rst
@@ -112,7 +112,10 @@ TechnicalLife
    represents the number of years that a technology operates before it is decommissioned.
 
 UtilizationFactor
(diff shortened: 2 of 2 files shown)


src/muse/constraints.py
```diff
@@ -530,7 +530,7 @@ def minimum_service(
         .drop_vars("technology")
     )
     capacity = convert_timeslice(
-        techs.fixed_outputs * techs.utilization_factor * techs.minimum_service_factor,
+        techs.fixed_outputs * techs.minimum_service_factor,
```
alexdewar, line 533: Where does the division by the maximum output happen?

alexdewar (approved):
LGTM, other than the one question I have.

src/muse/constraints.py
dalonsoa: It doesn't: the minimum service factor is the result of that division, that is why it is a fraction! `fixed_outputs` would be the absolute maximum and you multiply it by `utilization_factor` or the `minimum_service_factor` to recover the upper and lower limits of the production.

alexdewar: Oh I seeeee

tsmbland (approved):
Looks good to me!

tsmbland:
Did you mean to merge this into the `even-better-toml` branch?

dalonsoa:
No, my mistake :(

PR #392, opened 2024-07-02
Helper function for dropping timeslice

### LINKED ISSUE #363: Timeslice level names are no longer configurable [BUG]

tsmbland:
# Description

Adds a helper function for dropping the timeslice variable from data arrays, and uses this throughout the code instead of `drop_vars(["timeslice", "month", "day", "hour"])`

This partially fixes #363, but a few other changes will be required which I'm addressing in #381

## Type of change

## Key checklist

## Further checks


src/muse/timeslices.py
```diff
@@ -596,4 +596,15 @@ def represent_hours(
     return convert_timeslice(DataArray([nhours]), timeslices).squeeze()
 
 
+def drop_timeslice(data: DataArray):
```
dalonsoa, line 599: Ideally, add output type:
```suggestion
def drop_timeslice(data: DataArray) -> DataArray:
```

dalonsoa (changes requested):
Given that you're on it, add a dedicated tests that checks the new function.

src/muse/timeslices.py
```diff
@@ -596,4 +596,15 @@ def represent_hours(
...
+    """Drop the timeslice variable from a DataArray.
+
+    If the array doesn't contain the timeslice variable, return the input unchanged.
+    """
+    if "timeslice" not in data.dims:
+        return data
+
+    return data.drop_vars([c for c in data.timeslice.indexes])
```
alexdewar, line 607: Would this work:

```suggestion
    return data.drop_vars(data.timeslice.indexes)
```

If not, how about:

```suggestion
    return data.drop_vars(list(data.timeslice.indexes))
```

alexdewar (approved):
I've made a small suggestion (that probs won't work) but other than that, LGTM.

tsmbland:
@dalonsoa Would you mind taking a quick look at this so I can merge it?

dalonsoa (approved):
LGTM! Sorry, I didn't realized the changes had been implemented.

PR #394, opened 2024-07-02
Remove other references to SGI

alexdewar:
I noticed that there's still a reference to SGI in the copyright notice (you can see it in the footer of the documentation). I've updated the copyright owner to be "Imperial College London" and I changed the year while I was at it.

There was also a reference to SGI in `CITATION.cff` -- I think it refers to the old GitHub org, but I don't think that's needed: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

FYI @ahawkes

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -8,6 +8,6 @@ authors:
     - family-names: Hawkes
       given-names: Adam
 
-title: SGIModel/MUSE_OS
+title: MUSE_OS
 version: v1.0.1
 date-released: 2022-03-25
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,7 +6,7 @@
 # -- Project information -----------------------------------------------------
 
 project = "MUSE"
-copyright = "2022, Sustainable Gas Institute"
+copyright = "2024, Imperial College London"
 author = "Imperial College London"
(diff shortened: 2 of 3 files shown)


alexdewar:
I've also just noticed that the version number in `CITATION.cff` is out of date... shall I fix it and add it to the list in `.bumpversion.cfg`?

tsmbland (approved):

alexdewar:
I've also just noticed this URL in `pyproject.toml`: https://www.imperial.ac.uk/muse-energy/what-is-muse-/

If you try to go there, you get a 403 error. 

Maybe we should send users here instead: https://www.imperial.ac.uk/muse-energy/what-is-muse-/

alexdewar:
(Pinging @ahawkes re the URL -- this isn't urgent though)

ahawkes:
Yes the updated url that works would be good! Thanks

alexdewar:
Ok cool. I'll update it then.

Oddly, this Homepage field doesn't seem to be picked up by pip (even though I think we've done it correctly): https://github.com/pypa/packaging-problems/issues/606

But I think that's a problem for another day...

PR #395, opened 2024-07-03
Remove stray files in docs folder

alexdewar:
In the docs folder, there is a random empty file called "python", one called "objects.inv" (a Sphinx build artifact) and an unused "requirements.txt" file (we've since moved to pyproject.toml). Remove them.


dalonsoa (approved):
This is the oddest review I've seen - just deleted files :P

PR #398, opened 2024-07-03
Change column titles and order

tsmbland:
# Description

In line with the discussion [here](https://github.com/EnergySystemsModellingLab/MUSE_2.0/issues/101), I'm changing some of the column titles to be clearer. I've also made the column order more consistent between tables.


src/muse/data/example/default_new_input/agent_pairs.csv
alexdewar, line 1: I think I may have missed something, but what are agent pairs about?

src/muse/data/example/default_new_input/agents.csv
```diff
@@ -1,3 +1,3 @@
-agent,description,type,maturity_threshold,annual_cost_limit,search_rule,decision_rule
+agent_id,description,type,maturity_threshold,annual_cost_limit,search_rule,decision_rule
```
alexdewar, line 1:
```suggestion
id,description,type,maturity_threshold,annual_cost_limit,search_rule,decision_rule
```

src/muse/data/example/default_new_input/commodities.csv
```diff
@@ -1,4 +1,4 @@
-commodity_name,description,type,unit
+commodity_id,description,type,unit
```
alexdewar, line 1:
```suggestion
id,description,type,unit
```

alexdewar (approved):
There are a few columns which I think should be just `id`. I'd request changes but I'm going away and don't want to block merging.

src/muse/data/example/default_new_input/agents.csv
tsmbland: Well spotted, thanks!

src/muse/data/example/default_new_input/agent_pairs.csv
tsmbland: This is because agents have typically been made up of new and retrofit shares, so this is a way to link them together. I might change this later.

PR #401, opened 2024-07-04
Fix error message in `scipy_match_demand`

tsmbland:
I just realised that one of the changes I made in #381 broke an error message in the `scipy_match_demand` function. This fixes it

### DIFF
--- a/src/muse/investments.py
+++ b/src/muse/investments.py
@@ -310,7 +310,7 @@ def scipy_match_demand(
 
     from muse.constraints import ScipyAdapter
 
-    df_technologies = technologies.to_dataframe()
+    df_technologies = technologies.to_dataframe().reset_index()
 
     if "timeslice" in costs.dims and timeslice_op is not None:
         costs = timeslice_op(costs)
--- a/src/muse/investments.py
+++ b/src/muse/investments.py
@@ -310,8 +310,6 @@ def scipy_match_demand(
 
     from muse.constraints import ScipyAdapter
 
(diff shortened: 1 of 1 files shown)


tsmbland:
Or not...

dalonsoa:
I'm not sure I understand. Except for the `xfail` test, that PR was passing all checks, so what was broken?

tsmbland:
It's [this error message](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/06d35aff4de6c31dff44887ca01cb43f09a7b682/src/muse/investments.py#L345), which is raised when the solver can't find a solution. It's not part of the testing suite, so I only noticed when dealing with an infeasible model. Obviously my first attempt to fix it hasn't worked (although strangely it works for me locally), so I'll try something else

tsmbland:
Ok now it looks better

src/muse/investments.py
```diff
@@ -310,8 +310,6 @@ def scipy_match_demand(
 
     from muse.constraints import ScipyAdapter
 
-    df_technologies = technologies.to_dataframe()
```
dalonsoa, line 313: So... is this not needed anymore?

dalonsoa (approved):
Assuming the failing windows check isn't relevant, it looks good!

src/muse/investments.py
tsmbland: Nope. Previously it was converting the xarray to a dataframe and then getting the list of technologies from the dataframe, but you can get this directly from the xarray

PR #405, opened 2024-07-09
Tidy `find_equilibrium` function

tsmbland:
# Description

See #406 for more background info about this function, but basically this function has a load of calculations which aren't actually used, which makes it very hard to see what it's actually doing. It's also generally a bit of a mess, with a whole load of `if` statements and `break`s which can be tidied up.

I think there's more work that needs to be done on this function (see #407 for a start), which may well involve properly implementing some of the stuff that I've deleted here, but it will make our lives much easier if we have a cleaner starting point.

## Type of change

## Key checklist

## Further checks


src/muse/mca.py
```diff
@@ -538,94 +538,44 @@ def find_equilibrium(
...
-        if equilibrium_reached:
-            converged = True
-            new_price = prior_market["prices"].sel(year=market.year[1]).copy()
-            new_price.loc[dict(commodity=included)] = market.updated_prices.sel(
-                commodity=included, year=market.year[1]
-            )
-            market["prices"] = drop_timeslice(
-                future_propagation(  # type: ignore
-                    market["prices"], new_price
-                )
```
dalonsoa, line 579: I'm all in with dropping unnecessary code, but I think we should first understand well what this future propagation was supposed to be doing. And if it is doing something important but such a task is not really exploited because of an issues somewhere else, then let's fix that somewhere else.

I personally don't remember what this was supposed to be doing, to be honest.

tsmbland: Good point. Not sure what the intention was here, but I will look into it and, if necessary, deal with it in #407.

The main goal of this PR is just to clarify what the function is actually doing by removing the code that isn't used. My intention is to make all the necessary fixes in #407, and then we can merge them both together when done.

src/muse/mca.py
```diff
@@ -538,94 +538,44 @@ def find_equilibrium(
...
+    while iteration < maxiter and not converged:
         prior_market, market = market, prior_market
         market.consumption[:] = 0.0
         market.supply[:] = 0.0
         market, equilibrium_sectors = single_year_iteration(market, sectors)
 
-        check_demand_fulfillment(market.sel(commodity=included), tol_unmet_demand)
+        if maxiter == 1 or not equilibrium:
+            converged = True
+            break
```
alexdewar, line 558: So this means that if the maximum number of iterations is 1, then we treat it as if it's always converged. Is that right? Does `maxiter == 1` have a special meaning?

alexdewar (commented):
I have one (not particularly important) stylistic suggestion: I'd drop the `converged` variable and turn the loop into a normal `for` loop. I'd do the convergence check as one `if` statement inside the loop, returning from the function early if it converges (you might need to factor some bits out into a separate function to achieve this).

Anyhoo, this is very much secondary to fixing the rather disturbing issue(s) you've found :worried:

src/muse/mca.py
tsmbland: Yes, `maxiter = 1` is treated the same as `equilibrium = False`. See [here](https://muse-os.readthedocs.io/en/latest/inputs/projections.html#initial-market-projection)

src/muse/mca.py
tsmbland: Short answer: It doesn't seem to make a difference to the results whether I use the `future_propagation` function or not.

Long answer: It looks like the point of this function is to only update prices for "future" years (and only if the change is sufficiently large). For example, if you're simulating the period 2020-2025, this would ensure that only the 2025 price is updated by the market. Using it here it would only have an impact inside the `find_equilibrium` function loop, it's already applied to the output of this function [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/51f97ad485d2948b50a15282f9c8073dfa6d4f5d/src/muse/mca.py#L358). I'm still trying to figure out how prices are actually used in the model, so can't comment much more at the moment.

tsmbland: The other question is whether we want to include the price "buffering" which is coded up here but unused. 

The model calculates new commodity prices [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/51f97ad485d2948b50a15282f9c8073dfa6d4f5d/src/muse/sectors/sector.py#L329) based on supply and LCOE. 
As it stands, this is used directly to update prices in the market. However, there's a load of code in this function that calculates an 80/20 average between the original prices and the new prices, presumably with the intention of buffering price changes (although this average value doesn't get used). Is this buffering just an artefact of an old approach that we no longer want to do (in which case we delete the code and nothing will change), or is this something that we __do__ want to do (in which case we need to fix the implementation so the 80/20 average is actually used) @ahawkes do you have any idea about this?

ahawkes: Overall - the idea was that commodity prices are calculated in each iteration (for each commodity. time slice, region) - and then those updated prices are used in the next iteration to calculate things like LCOE etc. So market commodity prices influence investment decisions, and vice versa investment decisions influence commodity prices. Then a few ideas were implemented to get things to converge.

The 80/20 thing came from conversations with the US Energy Information Administration, where they have a similar model. They suggested this to stop price (and commodity supply/demand) jumping around too much. Looks like someone tried this and maybe gave up?

I don't think we can get to the bottom of this by end July, It would be good to know how the market prices are being calculated (is it weighted-average LCOE of all technologies contributing to a commodity's output?). And re convergence, are we checking for convergence of both prices and commodity quantities? Does this need to converge for every commodity and for every time slice, region?

src/muse/mca.py
alexdewar: Ah ok. Can we be sure that the user hasn't set `equilibrium = True` and `maxiter = 1` here?

src/muse/mca.py
tsmbland: Ok, good to know about the 80/20 conversations. I'll leave this out of the code for now but can potentially revisit later.

Re price calculations: Prices are calculated using [this function](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/51f97ad485d2948b50a15282f9c8073dfa6d4f5d/src/muse/quantities.py#L552). It looks like a weighted average LCOE of sorts. However, I'm not sure why the normalisation factor is summed over the timeslice dimension (which ends up making prices much lower than I'd otherwise expect). Am I missing something?? @dalonsoa any thoughts?

Re convergence: See [this function](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/51f97ad485d2948b50a15282f9c8073dfa6d4f5d/src/muse/mca.py#L663). It's checking either demand (specifically the _gap between supply and demand_, not demand itself) or prices, depending on the `equilibrium_variable` specified in `settings.toml`. If the quantity changes by less than the specified `tolerance` (for all timeslices, regions and commodities), then the MCA terminates. As mentioned in #415, the tolerance is absolute rather than relative (despite what the paper and documentation say).

dalonsoa: > Re price calculations: Prices are calculated using [this function](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/51f97ad485d2948b50a15282f9c8073dfa6d4f5d/src/muse/quantities.py#L552). It looks like a weighted average LCOE of sorts. However, I'm not sure why the normalisation factor is summed over the timeslice dimension (which ends up making prices much lower than I'd otherwise expect). Am I missing something?? @dalonsoa any thoughts?

Yep, this is exactly what it is:

```math
LCOE(ts) =\frac{Prod(ts)}{\sum_{ts} Prod(ts)}LCOE
```
The point is that the LCOE is the cost per annum, so the price per timeslice needs this kind of manipulation. Unless I'm missing something.

dalonsoa: Mmmm... indeed, that sounds about right.

src/muse/mca.py
tsmbland: I guess in that case the `maxiter=1` will win out. This is sort of explained in the documentation (albeit subtly):

"If the MCA works in an exogenous mode (setting market keyword equilibrium = false or maximum_iterations = 1), it will use the initial market projection as the projection for the the base year and all the future periods of the simulation"

The key word being "or"

src/muse/mca.py
```diff
@@ -538,94 +538,44 @@ def find_equilibrium(
...
 
-        equilibrium_reached = check_equilibrium(
+        # Check convergence criteria
+        check_demand_fulfillment(market.sel(commodity=included), tol_unmet_demand)
```
dalonsoa, line 561: This function returns `True` or `False`, but it is not assigned or used for anything, so it seems to me there's something missing here. As this was the behaviour before, let's leave it as it is, but seems odd to use this just to log the problematic behaviour.

dalonsoa (approved):
As all seems to be working and we have discussed the possible desired functionality, I think this is ready to go!

src/muse/mca.py
tsmbland: Yes I agree. See #410. It's also not very noticeable when the warning does get logged as it gets hidden among the barrage of other log messages. I'll leave it as it is for now, but might be something to revisit in the future.

PR #407, opened 2024-07-10
Fix `find_equilibrium` function

### LINKED ISSUE #406: `find_equilibrium` function always terminates on the second iteration [BUG]
I've been looking at the `find_equilibrium` function in `mca.py` to try and understand what the  `maximum_iterations` parameter is doing, as discussed in #372.

This function contains a loop which iteratively solves each sector (via the `single_year_iteration` function) and adjusts the price of each commodity accordingly. It then checks whether an equilibrium has been reached, and if not, moves to the next iteration to solve each sector again with the new commodity prices. The loop will finally terminate once an equilibrium is reached (i.e. when the change in supply and demand from iteration-to-iteration falls below a threshold). The `maximum_iterations` parameter defines the maximum number of times that this loop is allowed to run without converging.

Well I think that's the intention, but the function is one big mess.

To start with, there's a whole load of price manipulations within the function (e.g. [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/2e8e1b5d8d93317b4c6be361deaa43691606eaf0/src/muse/mca.py#L586)) which don't appear to be doing anything and just get overwritten by the `market["prices"] = drop_timeslice(market.updated_prices)` line at the top of the loop. I think possibly the intention here was to buffer the iteration-on-iteration price changes, but I'm not sure if this is really necessary (and in any case, it's not being used).

But I think there's a much bigger problem - it seems that loop is guaranteed to terminate on the second iteration (at least when `equilibrium_variable` is "demand" - I think when `equilibrium_variable` is "prices" it will always terminate on the third iteration (maybe)).

To make this clearer, I've written a tidier version of the function ([here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/c963227bcb9abbe829979cfbb0bea4b2b25adb01/src/muse/mca.py#L507)) which removes all of the unused computations, but should have the same behaviour (the tests are passing, at least). Looking at this, I think the problem is to do with the `prior_market, market = market, prior_market` line at the top of the loop. This sets `prior_market` to the `market` solved from the previous iteration (which is what we want), but it also reverts `market` back to the previous previous value (confused?), which then gets solved. The net result is that in the second iteration you end up comparing two markets that have been solved once with the same prices, so the convergence criteria always passes and the loop always terminates.

Just to convince yourselves further, try changing `maximum_iterations` for any model from 100 to 2, and you should get identical results.

I think a simple solution would be to change `prior_market, market = market, prior_market` to `prior_market = market.copy(deep=True)`. There may be more efficient solutions, but this seems to work and not always terminate on the second iteration.

I also think the convergence checks need another look, but let's get the loop working first.

tsmbland:
# Description

There are a number of issues with the `find_equilibrium` function, some of which are described in #406.

#405 is concerned with tidying up the function so we have a cleaner starting point to work off.

This PR is about fixing the bugs.

Changes (please see the `mca.py` file): 
- change the way that `prior_market` is created. This fixes the main problem where the loop would always terminate on the second iteration.
- update prices _before_ checking convergence. This will be essential when using "price" as the convergence criteria, so that you're properly comparing the price of the current market with the prior market.
- Move the demand fulfillment check outside the loop. The purpose of this is to raise a warning if demand isn't fulfilled, but it only makes sense to do this check on the final market. (See also #410)
- Proper use of the `future_propagation` function, to ensure that only future prices are changed. Note that this doesn't change any of the results, as agents are always looking at future prices to make investment decisions, but it's worth including anyway to ensure internal consistency.

Notes: 
- It looks like the original function was trying to buffer price changes from iteration-to-iteration (i.e. the price for the next iteration would be an 80/20 weighted average between the old price and the new price), although these prices were then overwritten. This was based on conversations with the US Energy Information Administration, see [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/405#discussion_r1682600171), but was only half-implemented. We're not going to include this for now, but may revisit in the future.

Fixes #406

## Type of change

## Key checklist

## Further checks


tsmbland:
I've merged this into the documentation branch, so you can see how this affects the tutorial model results [here](https://muse-os.readthedocs.io/en/documentation/user-guide/index.html)

dalonsoa (approved):
These changes make sense, the code is way simpler than originally and, I would assume, the numerical result still make sense, right?

PR #409, opened 2024-07-10
Quick fix for float-integer error

### LINKED ISSUE #408: Crash when some inputs are integers not floats
## Describe the bug

Muse sometimes crashes with the following error

numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

Past users have found that some inputs need to be "0." instead of "0" in order for muse to run, a tendency that may be associated with this problem.
 
## To Reproduce

The attached input files reproduce this error on the develop version at the time of writing.
[MuseInputFiles.zip](https://github.com/user-attachments/files/16163454/MuseInputFiles.zip)

## Context

- Operating system (eg. Windows 10): OSX
- MUSE version (eg. 1.0.1): develop
- Installation method (eg. pipx, pip, development mode): development mode
- Python version (you can get this running `python --version`): 3.9.18

martinstringer:
# Description

There is an instance of the *= operation in timeslices.py where the variables are not necessarily the same type. This sometimes causes the error describes in #408.

Fixes #408

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/timeslices.py
+++ b/src/muse/timeslices.py
@@ -549,7 +549,7 @@ def convert_timeslice(
         index = index.set_names(f"finest_{u}" for u in index.names)
         mindex_coords = xr.Coordinates.from_pandas_multiindex(index, "finest_timeslice")
         finest = finest.drop_vars(list(finest.coords)).assign_coords(mindex_coords)
-        proj0 *= finest
+        proj0 = proj0*finest
         proj0 = proj0 / proj0.sum("finest_timeslice")
     elif quantity is QuantityType.INTENSIVE:
         proj1 = proj1 / proj1.sum("finest_timeslice")


src/muse/timeslices.py
```diff
@@ -549,7 +549,7 @@ def convert_timeslice(
         index = index.set_names(f"finest_{u}" for u in index.names)
         mindex_coords = xr.Coordinates.from_pandas_multiindex(index, "finest_timeslice")
         finest = finest.drop_vars(list(finest.coords)).assign_coords(mindex_coords)
-        proj0 *= finest
+        proj0 = proj0 * finest
```
dalonsoa, line 552: Ok... but why this works and `*=` doesn't when the types are different?

martinstringer: My limited understanding was that a *= b is more memory efficient as it doesn't create a copy of a in the process. However, if a and b are different types (e.g. int vs float) it can't process. Meanwhile, a = a*b is less efficient but can work with different types because it can convert when it copies a to carry out the multiplication.

However, this may be wrong as it is based on reading between the lines of moderately-helpful documentation.

PR #412, opened 2024-07-12
Fix small issues with consumption, supply and prices outputs

### LINKED ISSUE #416: Model crashes when all price projections are set to zero
Issue raised by Alaa. Will add more info when I start investigating the issue

tsmbland:
# Description

Just a few small changes to fix some minor issues with the capacity, consumption, supply and prices outputs. See `src/muse/outputs/mca.py`

CHANGES:

The prices output contained a lot redundant code which has been removed:
- dropping empty rows is already dealt with by the `round_values` decorator
- the `keep_columns` functionality is already dealt with by the `standardize_quantity` operator (applied to the `to_csv` function)

Also with the prices output:
- the timeslice column in the output files was pretty ugly. This is now improved, and in line with the capacity and supply files
- I've also changed to `index` parameter for this output in all the `settings.toml` files to `false`. I was hoping to make this the default, as I can't see why anyone would want the index column in their output files, but I couldn't figure out how to do this.

The consumption and supply outputs were running into issues with the `keep_columns` functionality, as some of the columns were multi-index columns rather than regular columns. Applying `reset_index()` fixes that

Also, I've removed `reset_index()` from the capacity files to get rid of the "index" column 

Fixes #416 

## Type of change

## Key checklist

## Further checks


tsmbland:
Converting to draft again as not quite finished

src/muse/outputs/mca.py
```diff
@@ -151,50 +152,25 @@ def consumption(
     market: xr.Dataset, sectors: list[AbstractSector], **kwargs
 ) -> xr.DataArray:
     """Current consumption."""
-    from muse.outputs.sector import market_quantity
-
-    return market_quantity(market.consumption, **kwargs)
+    return market_quantity(market.consumption, **kwargs).to_dataframe().reset_index()
```
dalonsoa, line 155: We are not using type checking that much - if anything -, but since you're changing the return type, you should also update the signature of all of this functions to `pd.Dataframe`.

src/muse/outputs/mca.py
```diff
@@ -243,8 +219,6 @@ def sector_capacity(sector: AbstractSector) -> pd.DataFrame:
 
     capacity = pd.concat([u for u in capa_sector])
     capacity = capacity[capacity.capacity != 0]
-
-    capacity = capacity.reset_index()
```
dalonsoa, line 247: Why don't we need to reset index in this one?

dalonsoa (approved):
Just a couple of suggestions. 

I believe you can make the `index=False` option the default in this line:
https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/4fdf10b6e2d04d921c0536fcfba574fc53f3fd75/src/muse/outputs/sinks.py#L239

If you do, obviously you do not need to edit every single TOML file.

docs/tutorial-code/7-min-max-timeslice-constraints/2-max-constraint/settings.toml
```diff
@@ -17,6 +17,7 @@ tolerance_unmet_demand = -0.1
 quantity = "prices"
 sink = "aggregate"
 filename = "{cwd}/{default_output_dir}/MCA{Quantity}.csv"
+index = false
```
alexdewar, line 20: I think the `index` key is ultimately bundled up with other arguments passed to `to_csv`. If I'm right, this seems to be the only place where we do that: https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/36f8c778312e2969d4bfc7730646335fba161a91/src/muse/outputs/sinks.py#L246

alexdewar: So adding something like this might work:

```py
if not "index" in params:
    params["index"] = False
```

alexdewar (approved):
LGTM. I've made a suggestion about how to make `index == True` the default, which I think would be cleaner.

alexdewar:
(Sorry, I meant `index == False`!)

src/muse/outputs/mca.py
tsmbland: Because doing so creates a column called "index" in the output file. I guess `reset_index(drop=True)` would also work

tsmbland: It isn't needed here anyway. The other outputs need it because they have a multi-index with "month", "day" and "hour", so `reset_index` is required to create the respective columns, whereas here it's just an integer index.

PR #418, opened 2024-07-19
Fix prices calculation

### LINKED ISSUE #417: Incorrect weighted averaging in prices calculation
We have LCOE and production values for all assets and all timeslices, and we want to calculate a weighted average LCOE for each timeslice, based on production of the assets. I think this should be:

```math
LCOE(ts) =\frac{\sum_{assets} Prod(ts,assets) LCOE(ts,assets)}{\sum_{assets} Prod(ts,assets)}
```

But it's actually doing this:

```math
LCOE(ts) =\frac{\sum_{assets} Prod(ts,assets) LCOE(ts,assets)}{\sum_{ts,assets} Prod(ts,assets)}
```

Which I don't think is correct (if you're summing the numerator over assets, then you shouldn't be summing the denominator over both assets and timeslices)

_Originally posted by @tsmbland in https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/405#discussion_r1682824385_

tsmbland:
# Description

Fixes an incorrect calculation in the `supply_cost` function which is used to determine commodity prices

Also fixing the test, which looks like it was attempting the correct calculation, but gave up in favour of the incorrect calculation

Only regenerating some of the results files at the moment (as these will likely have conflicts with other PRs), so this won't pass the regression tests just yet

Fixes #417 

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Looks good to me! Simpler - also the tests!

PR #420, opened 2024-07-19
Fix issues with costs outputs

### LINKED ISSUE #383: LCOE output doesn't work [BUG]

tsmbland:
# Description

Fixes several issues with the costs outputs in `outputs/mca.py`. Some of which emerged after we upgraded pandas/xarray, some of which were issues before (but never caught). Will add more detail

Fixes #383 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/outputs/mca.py
+++ b/src/muse/outputs/mca.py
@@ -770,14 +770,13 @@ def sector_fuel_costs(
             data_agent["category"] = a.category
             data_agent["sector"] = getattr(sector, "name", "unnamed")
             data_agent["year"] = output_year
-            if len(data_agent) > 0 and len(data_agent.technology.values) > 0:
-                data_sector.append(data_agent.groupby("technology").fillna(0))
+            data_agent = multiindex_to_coords(data_agent, "timeslice").to_dataframe(
+                "fuel_consumption_costs"
+            )
+            if not data_agent.empty:
+                data_sector.append(data_agent)
     if len(data_sector) > 0:
-        output = pd.concat(
-            [u.to_dataframe("fuel_consumption_costs") for u in data_sector], sort=True
-        )
-        output = output.reset_index()
-
+        output = pd.concat(data_sector, sort=True).reset_index()
     else:
         output = pd.DataFrame()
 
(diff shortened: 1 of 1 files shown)


src/muse/outputs/mca.py
```diff
@@ -770,14 +770,13 @@ def sector_fuel_costs(
             data_agent["category"] = a.category
             data_agent["sector"] = getattr(sector, "name", "unnamed")
             data_agent["year"] = output_year
-            if len(data_agent) > 0 and len(data_agent.technology.values) > 0:
-                data_sector.append(data_agent.groupby("technology").fillna(0))
+            data_agent = multiindex_to_coords(data_agent, "timeslice").to_dataframe(
+                "fuel_consumption_costs"
+            )
+            if not data_agent.empty:
+                data_sector.append(data_agent)
```
dalonsoa, line 777: This looks ok, but it is doing something different than before. I assume this is correct, but I'm not sure why it is.

dalonsoa: ah, ok, you are combining it with what it was being done afterwards. OK, looks fine.

dalonsoa (approved):
Looks good. The code is way more readable now, in addition to fixing the bugs, which is always good news.

alexdewar:
FYI @tsmbland I'm just going through your PRs now

alexdewar (approved):
I've only given it a quick scan, but LGTM!

PR #428, opened 2024-07-25
Pin bottleneck dependency

tsmbland:
Tried to upgrade Sharwari's environment for the new MUSE version and had some problems. Turns out `bottleneck` also needs to be pinned


alexdewar (approved):
LGTM!

PR #432, opened 2024-07-26
Apply default constraints to all example models

tsmbland:
# Description

Some of the example models were missing the `minimum_service` and `demand_limiting_capacity` constraints, which are now default and should always be applied unless there's a very good reason not to.


src/muse/data/example/trade/settings.toml
```diff
@@ -50,11 +50,13 @@ agents = '{path}/technodata/residential/Agents.csv'
 existing_capacity = '{path}/technodata/residential/ExistingCapacity.csv'
 lpsolver = "scipy"
 constraints = [
+    # Optional, defaults to the constraints below
```
alexdewar, line 53: Something you sometimes see in example config files is for the default options to be present, but commented out (i.e. in this case `constraints`). Something to think about?

alexdewar (approved):
As long as it works, I'm happy!

PR #433, opened 2024-07-26
Update documentation for `demand_share` parameter

### LINKED ISSUE #341: Documentation for `settings.toml`: options for `demand_share` are out of date
The documentation for `settings.toml` ([here](https://muse-os.readthedocs.io/en/latest/inputs/toml.html)) lists the options for `demand_share` as:
- `new_and_retro`
- `market_demand`

According to the code, the valid options are actually:
- `new_and_retro` (or `default`)
- `standard_demand`
- `unmet_demand`

The documentation needs to be updated, including descriptions for each of the new options

tsmbland:
# Description

The options have changed since the documentation was last updated, as well as the default value.

Also, whilst #349 changed the default value to "standard_demand" in the sense that if `demand_share` isn't specified in the settings file it will default to this value, if you wrote `demand_share = "default"` it would still default to "new_and_retro". So I've fixed that.

Close #341

### DIFF
--- a/docs/inputs/toml.rst
+++ b/docs/inputs/toml.rst
@@ -522,13 +522,13 @@ to define the timeslice simply by referring to the slices it will use at each le
         in the simulation. There are currently two options:
 
         - :py:func:`~muse.demand_share.standard_demand` (default): The input demand is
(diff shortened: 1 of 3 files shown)


alexdewar (approved):
LGTM!

PR #447, opened 2024-08-07
Rename threshhold to threshold

### LINKED ISSUE #301: [BUG] Fix spelling "threshhold" -> "threshold"
There are lots of places where the incorrect spelling "threshhold" is used. Annoyingly, some of those places are in the names of fields in config files, so if we blindly rename them we will break backwards compatibility. Accordingly, we should probably deprecated these misspelled fields and issue a warning if the user uses them and add fields for the correctly spelled variant.

tsmbland:
# Description

Renames the incorrect spelling "threshhold" in three parameters. I've also added a warning if a user attempts to use "asset_threshhold" in the settings file. This isn't required for the other two parameters, see [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/301#issuecomment-2273806394)

Apparently none of these parameters are explained in the documentation, so there's nothing that needs changing in the documentation, but this obviously isn't a good thing!

Fixes #301

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/agents/agent.py
+++ b/src/muse/agents/agent.py
@@ -113,13 +113,13 @@ class Agent(AbstractAgent):
         objectives: Optional[Callable] = None,
         decision: Optional[Callable] = None,
         year: int = 2010,
-        maturity_threshhold: float = 0,
+        maturity_threshold: float = 0,
         forecast: int = 5,
         housekeeping: Optional[Callable] = None,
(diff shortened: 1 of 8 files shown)


alexdewar (approved):
LGTM! Good idea to warn users if they use the old option name.

PR #448, opened 2024-08-08
Remove ProcessName column from consumption files

### LINKED ISSUE #391: Remove `ProcessName` column from consumption files
See for example [this file](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/develop/src/muse/data/example/default/technodata/preset/Residential2020Consumption.csv)

It's really confusing having the `ProcessName` column in this file as demand is for the commodities (i.e. heat in this case) and not the technology. I'm pretty sure this column isn't used in the simulation, so should be safe to delete, but currently it has to be there or MUSE will complain

tsmbland:
# Description

One of my biggest gripes with the input files was that when specifying consumption data you had to specify a process (with the "ProcessName" column). This gives the impression that demand is for a particular process, but this is NOT true. Demand is for the commodities, and is agnostic to the process. Even though process must be specified in the file, this information isn't used in the model. If consumption data is specified for multiple processes, this just ends up being summed. Same thing with preset supply data which shares the same reader function (although this isn't used in any of the examples)

I think it would be much clearer to remove this column from the consumption files (or at least not to mandate it). However, we still need things to work if the column is present.

The main changes I've made are to the `read_csv_outputs` function (which I've also renamed), taking the summing operation from the `PresetSector` class and moving it here, and then dropping the ProcessName column. This will now work with or without the ProcessName column, and I think it makes it clearer what's actually going on.

I've also updated the documentation accordingly.

Fixes #391 

## Type of change

## Key checklist

## Further checks


src/muse/readers/csv.py
```diff
@@ -816,10 +816,19 @@ def expand_paths(path):
     datas = {}
     for path in allfiles:
         data = pd.read_csv(path, low_memory=False)
+        assert all(u in data.columns for u in indices)
+
+        # Legacy: drop ProcessName column and sum data (PR #448)
+        if "ProcessName" in data.columns:
```
alexdewar, line 822: How about warning users that having a `ProcessName` column is deprecated here?

src/muse/readers/csv.py
```diff
@@ -816,10 +816,19 @@ def expand_paths(path):
     datas = {}
     for path in allfiles:
         data = pd.read_csv(path, low_memory=False)
+        assert all(u in data.columns for u in indices)
```
alexdewar, line 819: Note that `assert` statements aren't included in optimised Python bytecode. But I guess if one of these columns is missing an error will be raised below?

alexdewar (approved):
LGTM. I would consider giving users a warning if they supply a `ProcessName` column, but other than that I think it's all good.

PR #450, opened 2024-08-12
Fix issue with sector outputs being called multiple times

### LINKED ISSUE #445: Values in sector output files duplicated when using "aggregate" sink [BUG]
Sector outputs (capacity, supply etc.) can either use the "csv" sink or the "aggregate" sink. The former creates separate csv files for each year in the simulation, whereas the latter combines these files to produce one overall file for all years (which for most purposes is a cleaner option).

However, there's a problem. When using the "csv" sink the files contain data for both the current and next year of the simulation (for example, the `2025.csv` file might contain data for 2025 and 2030, and the `2030.csv` file would contain data for 2030 and 2035). Then, because of this overlap, when combining this data using the "aggregate" sink, you end up with the values for every year being included twice.

Or something like that. The point is that when using the "aggregate" sink every row is duplicated in the output file, so if you then try and summarize/plot this data you're going to get values twice as high as expected.

tsmbland:
# Description

See [this comment](https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/445#issuecomment-2284022332). This PR splits the outputs into a separate function that get's called once for each sector after the MCA has converged, rather than every iteration as it was doing before.

Fixes #445 

## Type of change

## Key checklist

## Further checks


src/muse/sectors/sector.py
```diff
@@ -151,6 +151,8 @@ def __init__(
         It can be anything registered with
         :py:func:`@register_production<muse.production.register_production>`.
         """
+        self.output_data: xr.Dataset | None = None
```
alexdewar, line 154: Are you actually using this property before an `xr.DataSet` is assigned to it? If not, you could get away with this.

```suggestion
        self.output_data: xr.Dataset
```

alexdewar (approved):
Makes sense to me.

PR #451, opened 2024-08-13
Bump version to v1.2.0rc2

tsmbland:
Because there are a few more recent changes that weren't incorporated into rc1

Ideally we'd also merge #450 before creating the release. This would just mean that I can modify the [release guide](https://github.com/EnergySystemsModellingLab/MUSE_OS/discussions/429) to advise people to use the aggregate sink for sector outputs


tsmbland:
Also tempted to just make this v1.2.0, not bother with the whole release candidate thing, and just do more regular minor version updates. Thoughts? @alexdewar @dalonsoa

alexdewar:
> Also tempted to just make this v1.2.0, not bother with the whole release candidate thing, and just do more regular minor version updates. Thoughts? @alexdewar @dalonsoa

I think it doesn't matter *too* much either way, provided you communicate with the users so that they don't get bitten by breaking changes. Either way though, it would be helpful if users could actually run the new code to check we haven't broken anything we *don't* know about. Making an RC release might be a good way of communicating this: you can tell them that they've got an opportunity to tell you about this before you make a proper release and break their models. Up to you though.

PR #458, opened 2024-08-16
Modify `demand_limiting_capacity` constraint

### LINKED ISSUE #457: Problem with `demand_limiting_capacity` constraint for technologies that have multiple end-use commodities [BUG]
## Describe the bug

Let's say you have a technology that produces both petrol and diesel. The demand for petrol is 1, and for diesel it's zero (in all timeslices).

Because demand for diesel is zero, the `demand_limiting_capacity` constraint for this technology will be zero for diesel. Any investment in this technology to meet the petrol demand will cause the diesel constraint to be violated.

If this is the only technology that can produce petrol, the model will fail.

I'm pretty sure this is what's going on in Alaa's model.

tsmbland:
# Description

This modifies the demand-limiting capacity constraint to address the issue described in #457. I've added lots of comments to the code so hopefully everything is relatively clear, but happy to have a chat about it if not.

None of the example models are affected, as this only affects certain technologies like refineries that have multiple end-use commodities, but I've tested this with Alaa's Qatar model and the results look reasonable.

Fixes #457 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
     "License :: OSI Approved :: GNU General Public License v3 (GPLv3)"
 ]
 dependencies = [
-    "numpy>=2.0",
+    "numpy==2.0",
     "scipy>=1.13",
     "pandas>=2.2",
     "xarray>=2024.6",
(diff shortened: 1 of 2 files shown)


dalonsoa (changes requested):
Looks good but very hard to understand the underlying logic - specially in the long term. I suggest you extract the new code as a standalone function and add some dedicated test - can even be a doctest, so it is included with the documentation - illustrating what is going on here and what problem is solving.

tsmbland:
@dalonsoa I've extracted the code to a new function and created a doctest, as suggested, so hopefully it's clearer now.

Also, things started to fail with python 3.12 due to the numpy 2.1 release, so I've had to pin numpy to 2.0 for now.

PR #461, opened 2024-08-20
Use aggregate sink for all example models

tsmbland:
# Description

Since fixing the aggregate sink in #450, it makes sense to actually use it for the example models. This essentially just combines all the sector outputs into a single file for all years, rather than having separate files for each year with overlapping data between files (which I've always found pretty annoying). 

There are more lines deleted than added (always satisfying), which is mostly because we now no longer have data duplicated between files. The only other difference is that the previous files included capacity data for existing assets beyond the timeframe of the simulation, which is cropped out in the aggregate files. I don't imagine this being a huge issue though.

(Almost none of these files are actually used in the tutorials, and most users will find everything they need in the MCACapacity and MCASupply files, so I'm tempted to delete most of them anyway, but I'll deal with that another time)


dalonsoa (approved):
Another of those small changes that make everyone's life better. Good work!

Issue #288, opened 2024-05-03
Change logic of how LCOE is used to calculate commodity prices


dalonsoa:
Thanks. That's a good point, and could be an issue in some circumstances. 

The price calculation is important to determine internal commodity prices, which are then used in calculation of investment-related LCOE (and other objectives) of assets that consume those commodities, and so-on. The issue you describe would only be an issue where no commodity is produced by any technology in a time slice - which is very unlikely in a standard systems model - but could be an issue if the framework is used in unexpected ways (which is entirely possible).

For asset investment LCOE calculation, the change proposed here will make the LCOE just a single value that depends on annual utilisation (so essentially its the same in every time slice).

For the commodity price calculation (which presumably happens elsewhere in MUSE), where the commodity price is assumed to be based on LCOE (I think there are options here, it's either LCOE-based or marginal price based, where marginal price is variable costs only), there are a few defensible ways to do this. One could (a) use the output-weighted average LCOE of every asset contributing in the time slice, or (b) use the highest LCOE of all assets contributing in a time slice. Then we come to your question - if all assets have zero output of the commodity in a time slice, what do we do? I think we would need to break that down into a few cases, (1) where all of the available assets (including all possible new assets) have an upper bound of output in that time slice of zero, then price should be infinite because there's simply no way to get that commodity in that time slice, and (2) where there are available assets that could produce the commodity in that time slice (but aren't doing so yet), it perhaps should the LCOE or marginal cost of the cheapest one (though this is debateable - actual price will depend on which asset is actually chosen). Anyway, it seems like this price issue is not the one you're dealing with in relation to this bug - rather you are looking at the asset investment - and have found a serious issue there. Should a new issue be raised for this price calculation problem?

_Originally posted by @ahawkes in https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/246#issuecomment-2074924498_

tsmbland:
I think #609 reasonably addresses this

PR #462, opened 2024-08-21
Create costs module and simplify code for objectives and outputs

tsmbland:
# Description

I've been digging into the code to get some more clarity on how costs (LCOE, ALCOE, EAC etc.) are calculated and used, but the issue is that the code for calculating these costs is defined in multiple places which makes it really tricky to see what's going on, could lead to inconsistencies, and is just a bad idea in general.

To tidy things up, I've created a new `costs` module that houses these calculations, and modified other parts of the codebase (specifically the outputs and objectives modules) to call these functions rather than repeating the calculations. Hopefully this will also make #288 easier to tackle.

It looks like a massive PR, but it's almost entirely just moving and deleting code, and creating some new tests. 

Some more details about the specific changes:

Costs module:
- I've created functions for NPV, NPC, EAC and LCOE, which are derived from code that was previously in the `objectives` module
- I've moved `annual_levelized_cost_of_energy` and `supply_cost` here from `quantities.py`
- I think there's more work that could be done here, e.g. creating functions for fuel costs, capital costs etc. and calling these functions as appropriate from LCOE, EAC etc. But that can be done later.
- Tests for this module are in `test_costs.py`. At the moment, this just goes through each function with some example data to check that the function runs, and that the dimensions of the result are as expected. I notice there are some differences between functions (e.g. the ALCOE function gets rid of the `asset` dimension), but this is something that can be looked into later and potentially homogenized.
- It might be a good idea to do some more thorough tests to check the values returned by these functions in a few scenarios. Not sure the best way to do this to be honest because the example data is randomly generated so the values will change every time.
- Still a bit unsure about the difference between capital recovery factor and discount factor, so I've included functions for both of these, although they seem very similar 

Outputs module:
- Calls the new LCOE and EAC functions from the cost module, rather than repeating all the calculations
- I've moved `capacity_to_service_demand` (which was defined multiple times) to `quantities.py`

Objectives module:
- Modified the objectives to call the costs functions where appropriate
- I've also modified what's passed to the objectives. Previously, objectives would take five objects: `agent`, `demand`, `search_space`, `technologies` and `market`. In all cases, `search_space` would be used to filter the technologies and reduce the dimensions of the demand object. Agent was also used in the filtering (to filter objects based on the agent's region and forecast year), and to get the forecast year while computing the objectives. Market was only used to get commodity prices.
- All this filtering was taking place in the objectives themselves, which resulted in a lot of repeated code. I've simplified things so the filtering is done outside the objectives, and the objectives are only passed the data that they need (`technologies`, `demand` and `prices`). The objectives can get the forecast year from the demand object (a bit of a hack, but it works ok for now)
- I've re-written the tests for this module, which were a bit of a mess before. The aim now is just to go through each objective with some example data to check that they run without error, and that the dimensions of the result are as expected (this is also really useful to document the dimensions of each output, as there's some variability between objectives). The previous tests also checked the values, but they were essentially just copying and pasting code from the objectives themselves to compute the expected values, which seems a bit pointless. I also don't think this is the appropriate place to be checking the values, as (now) most of the objectives are just calling functions from the costs module, so if we're checking values this should be done in `test_costs.py`

## Type of change

## Key checklist

## Further checks


tsmbland:
Actually the results are now different when I use the `ALCOE`, `capital_costs` and `efficiency` objectives, so I need to investigate why that is...

tsmbland:
> Actually the results are now different when I use the `ALCOE`, `capital_costs` and `efficiency` objectives, so I need to investigate why that is...

I think there was an error before with the `ALCOE` objective, as it was returning data for multiple years, but I think the agent should only have objective data for the investment year (at least that's what happens with all the other objectives). So the results are different now with this objective, but I think it was wrong before.

Very confused about `capital_costs` and `efficiency`. If the model runs all the way to the end, then the results are no different compared to before (at least with the example I've been using). But if the model terminates early because the growth parameters aren't high enough, then it appears to terminate in a different place (I have absolutely no idea why) so the interim results that you get might be different. Weird, but I think it's all ok.

dalonsoa (approved):
I think this is a much needed improvement, reducing the redundancy of functions that were supposed to be doing the same thing. Homogenising the interface of the functions and reducing code duplication is also great. I think that any discrepancy with the previous results is likely to be the result of fixing a bug we didn't know we had. None of the tutorials or examples need updating so, at least for the functionality they use like LCOE, these changes have no negative impact. 

Testing the results of mathematical formulas in a meaningful way is always tricky, as the only way of actually test if the numerical values are correct is to calculate them, which in turns uses the formula you're trying to test. Testing the properties of the outputs rather than the numerical values seems like a good compromise.

If you want to take one more step, you could use the [hypothesis](https://hypothesis.readthedocs.io/en/latest/) package to evaluate the functions under a range of conditions, and then evaluate properties of the results rather than specific numerical values (eg. all should be positive values, or be within certain range, or add up to 1, etc.). This is useful to find edge cases that are difficult to find manually.

tsmbland:
Thanks. I'm actually quite keen to give hypothesis a try, and this seems like a good place for it, so I'll open up another issue for that

PR #483, opened 2024-09-18
Fix bugs in the carbon budget bisection algorithm

tsmbland:
# Description

This fixes a number of problems with the bisection method for setting the carbon price when using a carbon budget. 

One problem was with the calculation of initial price bounds. Firstly, the calculation of the current year in `bisection` was just wrong, which often led to `time_exp` being zero (so the upper and lower bounds would be the same). Secondly, it couldn't deal with an initial carbon price of zero (both bounds would then be zero). Thirdly, the bounds were way to narrow so unlikely to capture the solution. These are all fixed.

The main loop in the algorithm was also a mess, including the convergence criteria, so this has been tidied up and simplified (well, pretty much rewritten completely).

The `min_max_bisect` function was also a mess, so this has been tidied and comments added to explain the logic. The inverted bounds case seemed to be only partially implemented, so I've filled this in.

I've also created a `tolerance` parameter which allows the user to define how close emissions need to be to the budget for the loop to terminate (previously this was hard-coded at 10%). For example, adding to the settings file:

```toml
method_options.tolerance = 0.05
```

I've added a warning for users if the algorithm doesn't reach the target emissions for a particular year.

Finally, I think the flow of the algorithm was wrong. It would solve the market to find the prices of all the commodities, then adjust the carbon price with these prices fixed, and then solve again with the final carbon price fixed and all the other prices allowed to change. This meant that the final market won't necessarily be the same as the market settled on in the bisection algorithm, so emissions might be different. I've changed it now so all the prices are allowed to change together (i.e. every time the carbon price is changed, the other commodity prices are also allowed to change). The main code change to achieve this was to remove the `1` from the `new_market = equilibrium(...` line in `bisect_loop`, and also remove the the initial market solve from `update_carbon_price`. It might take longer to run now, but I think this is necessary.

It's possible that some of these changes might break the other carbon budget algorithm (`fitting`), and some of the improvements I've made here could also be applied to that algorithm, but I've opened up another PR (#484) to deal with that.

I've also added a new tutorial explaining how to use the carbon budget, but keeping that separate to make reviewing easier (#486). You can see the tutorial [here](https://muse-os.readthedocs.io/en/documentation/user-guide/carbon-budget.html)

This will become v1.2.1rc1 once it's ready, so I've gone ahead and bumped the version

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/carbon_budget.py
+++ b/src/muse/carbon_budget.py
@@ -527,7 +527,7 @@ def min_max_bisect(
 
     if ub > threshold and lb < threshold:
         # Threshold is between bounds -> perform bisection
-        midpoint = (low + up) / 2.0
+        midpoint = round((low + up) / 2.0, 2)
         m = bisect_loop(market, sectors, equilibrium, commodities, midpoint)
         emissions_cache[midpoint] = m
         if m < threshold:
@@ -540,7 +540,7 @@ def min_max_bisect(
     # Inverted bounds (i.e. increasing price leads to increasing emissions)
     # Unlikely case, but included for completeness
     if ub < threshold and lb > threshold:
-        midpoint = (low + up) / 2.0
+        midpoint = round((low + up) / 2.0, 2)
         m = bisect_loop(market, sectors, equilibrium, commodities, midpoint)
         emissions_cache[midpoint] = m
         if m > threshold:
--- a/src/muse/carbon_budget.py
+++ b/src/muse/carbon_budget.py
@@ -404,7 +404,7 @@ def bisection(
     """
     # We calculate the carbon price and emissions threshold in the forecast year
     future = market.year[-1]
-    threshold = carbon_budget.sel(year=future).values
+    threshold = carbon_budget.sel(year=future).values.item()
     price = market.prices.sel(year=future, commodity=commodities).mean().values
 
     # Initial lower and upper bounds on carbon price for the bisection algorithm
@@ -433,12 +433,6 @@ def bisection(
             )
         lb = emissions_cache[up]
 
-        # Exit loop if emissions are identical at both prices
-        # (i.e. changing the carbon price has no impact on emissions)
-        # In this case we'll take the lower price
-        if ub == lb:
-            return low
-
         # Exit loop if lower or upper bound on emissions is close to threshold
         if abs(ub - threshold) <= abs(tolerance * threshold):
             return low
@@ -459,8 +453,9 @@ def bisection(
             emissions_cache=emissions_cache,
         )
 
-    # If convergence isn't reached, take an average of the final bounds
-    return (low + up) / 2.0
+    # If convergence isn't reached, new price is that with emissions closest to
+    # threshold. If multiple prices are equally close, it returns the lowest price
+    return min(emissions_cache, key=lambda k: (abs(emissions_cache[k] - threshold), k))
 
 
 def min_max_bisect(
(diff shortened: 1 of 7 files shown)


src/muse/carbon_budget.py
```diff
@@ -385,104 +386,94 @@ def bisection(
...
-                new_price = up
+        ub = emissions_cache[low]
+        if up not in emissions_cache:
+            emissions_cache[up] = bisect_loop(
+                market, sectors, equilibrium, commodities, up
+            )
+        lb = emissions_cache[up]
+
+        # Terminate early if many consecutive emissions are the same
+        if len(emissions_cache) >= early_termination_count + 2:
```
dalonsoa, line 444: I'm not entirely sure I follow this.

Every time you have an `up` or `low` that is not in the cache, you add them to the cache. So, the cache grows when `up` and `low` are different. However, this is suppose of break the loop if they are the same (too many times). Then, shouldn't this be checking that the cache is not growing and not that is growing and becoming too large?

src/muse/carbon_budget.py
```diff
@@ -505,68 +497,58 @@ def min_max_bisect(
...
+        emissions_cache[midpoint] = m
+        if m > threshold:
+            # Midpoint price is too high -> becomes new upper bound
             up = midpoint
-
         else:
-            # midpoint is smaller than ub
-            # ub is a function of low price
+            # Midpoint price is too low -> becomes new lower bound
             low = midpoint
```
dalonsoa, line 552: I'd suggest to separate these 4 cases into separate functions to ease testing and debugging.

src/muse/mca.py
```diff
@@ -243,38 +243,14 @@ def update_carbon_budget(self, market: Dataset, year_idx: int) -> float:
...
-        sectors do, the carbon budget limit is not exceeded.
+        """Calculates the updated carbon price.
 
         Arguments:
-            market: Market, with the prices, supply, consumption and demand.
+            market: Market with the prices, supply, consumption and demand.
 
         Returns:
-            The new carbon price or None
+            The new carbon price. If None, the price is not updated.
```
dalonsoa, line 252: I don't think the carbon price can be `None` anymore. If it can, then the docstrings for the different carbon methods and their signature needs updating.

dalonsoa (changes requested):
I've a couple of comments but it mostly looks ok.

I think, however, that this could benefit from some tests. One of the reasons this was wrong and no one had noticed until now is that there were not tests checking that this was doing what is was supposed to be doing. It might be a good time to fix that.

src/muse/carbon_budget.py
tsmbland: The main reason I included this was for models where it's impossible to get emissions close to the budget (perhaps low-carbon technologies don't exist that can meet the demand). In this case, the upper bound price will keep being ramped up until a plateau on emissions is reached. If this plateau is not within range of the budget, the price will keep being ramped up until the maximum iterations is reached, so I needed a check in place to prevent this.

`up` and `low` refer to prices, but this is checking whether _emissions_ are the same too many times

src/muse/mca.py
tsmbland: I've addressed this in #484

tsmbland:
@dalonsoa I've added unit tests for individual components of the algorithm. I think the tutorial in #486 acts as a reasonable integration tests, at least to test that the algorithm runs to completion without error.

src/muse/carbon_budget.py
dalonsoa: I'm not sure I follow. Where are you checking that emissions are the same too many times? Here you just check that the emissions cache length is above certain threshold.

tsmbland: It's in the three lines below that `if` statement. The `if` statement is just to make sure the cache has enough values to check (and I thought it was reasonable to exclude the starting bounds from the check, hence the +2)

dalonsoa: Got it.

src/muse/carbon_budget.py
```diff
@@ -385,199 +387,221 @@ def bisection(
...
+        # Inverted bounds (i.e. increasing price leads to increasing emissions)
+        # Unlikely case, but included for completeness
+        method = bisect_bounds_inverted
+
+    else:
+        # lb_price_emissions or ub_price_emissions is equal to the target, so we can
+        # return the bounds unchanged
+        return lb_price, ub_price
+
+    return method(
```
dalonsoa, line 536: Neat!

dalonsoa (approved):
Looks great!

I agree. I think it is good to have more unit tests - as you have added - but the notebooks themselves serve well as integration tests.

PR #484, opened 2024-09-19
Refactor fitting algorithm, some global changes to carbon_budget module

tsmbland:
# Description

this change.*

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/carbon_budget.py
+++ b/src/muse/carbon_budget.py
@@ -7,10 +7,9 @@ from scipy.optimize import curve_fit
 
 from muse.mca import FindEquilibriumResults
 from muse.registration import registrator
-from muse.sectors import AbstractSector
 
 CARBON_BUDGET_METHODS_SIGNATURE = Callable[
-    [xr.Dataset, list, Callable, xr.DataArray, xr.DataArray], float
+    [xr.Dataset, Callable, xr.DataArray, list], float
 ]
 """carbon budget fitters signature."""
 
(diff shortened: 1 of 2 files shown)


src/muse/carbon_budget.py
```diff
@@ -138,72 +125,14 @@ def fitting(
...
-            carbon_price,
-            carbon_budget,
-            sample_prices,
-            new_price,
-            commodities,
-            price_too_high_threshold,
-        )
+    # Cap price between 0.01 and price_too_high_threshold
+    if refine_price:
+        new_price = min(new_price, price_too_high_threshold)
```
dalonsoa, line 130: Why is there so much difference between the original refine price function and this one, which just one line?

src/muse/mca.py
```diff
@@ -128,9 +128,9 @@ def __init__(
         carbon_price: Sequence | None = None,
         carbon_commodities: Sequence[str] | None = None,
         debug: bool = False,
-        control_undershoot: bool = True,
-        control_overshoot: bool = True,
-        carbon_method: str = "fitting",
+        control_undershoot: bool = False,
+        control_overshoot: bool = False,
+        carbon_method: str = "bisection",
```
dalonsoa, line 133: Probably not many people is using it, but as you are changing the defaults, make sure that this change is clearly explained in the next release notes.

dalonsoa (approved):
looks good! The documentation is much clearer and and, in general, the code is tidier. I am not sure I understand, though, why the `refine_new_price` function was 50 lines of code before and now is just 1 line. What has changed?

src/muse/carbon_budget.py
tsmbland: There was some additional logic in the original function that was looking at the growth rate of the carbon price, but it doesn't seem to be properly implemented and it's not clear what the intended behaviour was. There might be something to think about here, like putting an upper limit on the rate of price increase/decrease, but the documentation only talked about capping the price based on `price_too_high_threshold`, so this is what I went for

PR #486, opened 2024-09-19
Carbon budget tutorial

tsmbland:
# Description

Adds a new tutorial for using a carbon budget

You can see the tutorial [here](https://muse-os.readthedocs.io/en/documentation/user-guide/carbon-budget.html)


docs/user-guide/carbon-budget.ipynb
```diff
@@ -0,0 +1,179 @@
...
+    "\n",
+    "The other lines indicate settings used in the carbon budget algorithm, which modifies the carbon price so that carbon emissions come as close as possible to the budget. The method we are using is the [bisection method](https://en.wikipedia.org/wiki/Bisection_method), which iteratively halves a carbon price interval until emissions are within tolerance of the target. `sample_size` and `tolerance` are parameters used in the bisection algorithm: `tolerance` controls how close emissions need to be to the budget (e.g. 20%) for the algorithm to converge on a solution; `sample_size` is the maximum number of carbon prices that will be tested. Reducing `tolerance` and increasing `sample_size` can help achieve emissions as close as possible to the budget, but the simulation may take much longer to run.\n",
```
dalonsoa, line 42: The meaning you state here for the `sample_size` is the one for the `fitting` algorithm. For the `bisection` method, it is (taken from the docstrings) "Maximum number of iterations for bisection". I think that we should either use a different parameter (`sample_size` for `fitting` and `steps` for `bisection`, for example) or very clearly explain what `sample_size` represents depending on what algorithm you are using.

docs/user-guide/carbon-budget.ipynb
```diff
@@ -0,0 +1,179 @@
...
+    "\n",
+    "See [here](../inputs/toml.rst) for more details about carbon budget settings.\n",
```
dalonsoa, line 44: Can you link the specific section of that file with something like:
```suggestion
    "See [here](../inputs/toml.rst#carbon-market) for more details about carbon budget settings.\n",
```
I'm not sure if rst let you do that...

dalonsoa (approved):
Very nice! I just have a couple of minor comments.

docs/user-guide/carbon-budget.ipynb
tsmbland: Your suggestion worked, thanks

Issue #498, opened 2024-09-30
Output warnings to a file


tsmbland:
Currently MUSE produces a few types of warnings when issues arise. These have important information that users need to pay attention to, but currently they get buried in the deluge of messages that MUSE prints so are likely to be ignored

I think warning messages should be outputted to a file so they're easier to see, and to keep a permanent record. To be honest it's probably a good idea to have some kind of log file anyway including information like the MUSE version and Python version, so warning messages can be a part of this

PR #507, opened 2024-10-02
Simplify the agents module, and additional changes

tsmbland:
This PR combines a few small changes which I made while looking through the code and trying to understand how everything works. A lot of the changes are just to make the code more readable (e.g. adding inline comments), but there are a few changes to the actual code as well:

- Remove some default values for function arguments. Specifically things like the year and forecast length. I think generally it's better to be explicit about these things, and I couldn't really see any reason you'd ever want these to rely on the defaults (apart from maybe making the tests simpler)

- Remove `agents.factories.factory` (never used)

- Remove `demand_share.new_demand` (never used)

- Remove the `protected` argument from `cliff_retirement_profile`. The code was needlessly complex, and the equivalent can be done by clipping the input

- Removing some unnecessary interpolation operations from `Sector.next`

- Tidy up the `Agent` and `InvestingAgent` classes. I didn't like how a lot of the code for the `InvestingAgent` was wrapped up in the `next` method of the `Agent` class. I've tidied things up so that all the code related to investments is in the `InvestingAgent` class

- `aggregate_lp` no longer returns anything. It's very difficult to tell when this function is supposed to output something, and what that output was supposed to be used for. Looking through the code, it appears this would only ever be used when using a noninvesting agent (`Agent` class rather than `InvestingAgent`), which is used when specifying the agent type as "agent", "default" or "standard". However, the only options given for this parameter in the documentation are "new" and "retrofit" (which both use the `InvestingAgent` class), so I can only assume these are legacy options, and this pathway isn't required any more (although maybe I'm missing something?). In which case, we can probably simplify the agents module even more that I've done here, bit I'll leave it for now. For what it's worth, the "trade" example model _does_ use the "default" agent type, but isn't affected by these changes

This doesn't address any particular issues, but I'm increasingly finding that lack of readability is a barrier for bug-fixing and development, so I think any attempts to address this will be helpful in the long run, even if we end up breaking backwards-compatibility or undocumented features.

Probably should have broken this into multiple PRs, but I didn't really go into this with any sort of plan, so this is just how its ended up.


src/muse/demand_share.py
```diff
@@ -474,28 +471,30 @@ def _inner_split(
...
-        # Calculates the total demand assigned in the previous step with the "method"
-        # function across agents and assets.
         total: xr.DataArray = summed_shares.sum("asset")
     except AttributeError:
         raise AgentWithNoAssetsInDemandShare()
 
     # Calculates the demand divided by the number of assets times the number of agents
     # if the demand is bigger than zero and the total demand assigned with the "method"
     # function is zero.
-    unassigned = (demand / (len(shares) * len(summed_shares))).where(
```
tsmbland, line 498: Reeeeeally bad idea to call `len` on a DataArray (`summed_shares`), as this will just take the length of the first dimension, which isn't guaranteed to be the dimension you're after ("asset" in this case), unless you're really careful about it. In this case we get away with it, but this has bitten me on another PR and took me a long time to figure out.

Also, I still don't really understand what this function is doing. Why is it multiplying the number of agents by the number of assets? What does `totals` represent?

src/muse/demand_share.py
```diff
@@ -630,37 +638,19 @@ def new_and_retro_demands(
...
         - service
     ).clip(min=0)
-
     if "year" in retro_demand.dims:
         retro_demand = retro_demand.squeeze("year")
 
     return xr.Dataset({"new": new_demand, "retrofit": retro_demand})
-
-
-def new_demand(
```
dalonsoa, line 649: I remember to have coded this when removing the retrofit agents, and I guess I did it for a reason to be used somewhere, but it is true that it doesn't seem to be used at all. Well, bye bye function!

src/muse/investments.py
```diff
@@ -186,19 +187,13 @@ def cliff_retirement_profile(
...
     We could just return an array where each year is represented. Instead, to save
     memory, we return a compact view of the same where years where no change happens are
     removed.
 
     Arguments:
         technical_life: lifetimes for each technology
-        current_year: current year
-        protected: The technologies are assumed to be renewed between years
-            `current_year` and `current_year + protected`
+        investment_year: The year in which the investment is made
```
dalonsoa, line 196: The year index? It is not the actual year, as in 2025. And if it is, it should not default to zero - if defaulting to anything.

src/muse/sectors/subsector.py
```diff
@@ -51,62 +50,33 @@ def invest(
...
             )
 
+        # Agent housekeeping
         for agent in self.agents:
             agent.asset_housekeeping()
 
-        lp_problem = self.aggregate_lp(
-            technologies, market, time_period, current_year=current_year
-        )
-        if lp_problem is None:
```
dalonsoa, line 71: Yeah, I never understood the remaining of this function...

src/muse/utilities.py
```diff
@@ -141,30 +141,40 @@ def reduce_assets(
     """
     from copy import copy
 
+    assets = copy(assets)
```
dalonsoa, line 144: Is it copy enough and not necessary deepcopy?

tests/test_subsector.py
```diff
@@ -105,20 +104,7 @@ def test_subsector_noninvesting_aggregation(market, model, technologies, tmp_pat
         commodity=technologies.commodity, region=technologies.region
     ).interp(year=[2020, 2025])
     assert all(agent.year == 2020 for agent in agents)
-    result = subsector.aggregate_lp(technologies, market)
```
dalonsoa, line 108: Why was this returning something different that `None`? I think this test only makes sense if `aggregate_lp` returns something different than `None`, which we have seen cannot happen. And if it can happen, as this test was demonstrating, under what conditions? Are we removing functionality that is actually useful (not necessarily in use)?

dalonsoa (approved):
I've made a few comments and suggestions here and there, but everything seems to make sense to me. 

It is illustrative of the state of MUSE that after all the code you have removed or changed, the tests are largely the same.

tests/test_subsector.py
tsmbland: When using the `Agent` class, `aggregate_lp` would return the search space and constraints, and then these would be used to perform investments withing `Subsector.invest`. Conversely, the `InvestingAgent` performs investments within its own `next` method. In theory the net result should be the same, so I can't see any reason to use `Agent` over `InvestingAgent`. I'm guessing it's just a legacy thing that was never tidied up

src/muse/utilities.py
tsmbland: It doesn't seem to make a difference, but I'll switch to deepcopy to be on the safe side

src/muse/agents/agent.py
```diff
@@ -263,22 +294,75 @@ def next(
...
             return None
 
+        # Calculate the search space
         search_space = (
             self.search_rules(self, demand, technologies, market).fillna(0).astype(int)
         )
 
+        # Skip forward if the search space is empty
         if any(u == 0 for u in search_space.shape):
             getLogger(__name__).critical("Search space is empty")
```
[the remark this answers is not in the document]
tsmbland: In theory it can continue, the agent just won't perform any investments for that year. If this happens, then it probably means there's a mistake with the way the model is defined, so this is something the user should definitely pay attention to. I'm a bit concerned that a lot a important messages are getting ignored, so have suggested these get outputted to a file (#498)

PR #510, opened 2024-10-04
Delete legacy sector

### LINKED ISSUE #358: python -m pytest throws "10 skipped tests"
Should be tackle them individually, or, given they are skipped by default, get rid of them entirely?

### LINKED ISSUE #509: Delete legacy sectors
There's a fair bit of code to do with legacy sectors (e.g. `legacy_sector.py`, and the `calibrate_legacy_sectors` function in `mca.py`). I guess this is included to maintain backwards compatibility, but this code isn't covered by any tests so there's no guarantee it even still works.

If people want to use this feature (very unlikely I would say), then they will always have access to older versions of MUSE. Otherwise, I see no point keeping this code, and it would make my life easier to get rid of it and generally shrink the codebase

tsmbland:
Deletes the `legacy_sector` module, and some other parts of the code that were only used by the legacy sector.

Obviously this breaks backwards compatibility, but given how minimal the tests are (and they're not run as part of CI anyway), there's no guarantee that the legacy sector still works anyway.

This will just allow me to focus on the important parts of the code when making changes and not waste time fixing code that won't ever be used. If anyone still needs to use this, they will still have access to older version of MUSE

Obviously I'll bump the minor version once this is merged as it breaks backwards compatibility

Closes #509
Closes #358


dalonsoa (approved):
Wow! This is like breaking free! It looks good to me. 

I think you can also delete this one, but I'm not entirely sure: https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/8e8248627c6e16aa477657431b535d102ee9428b/tests/conftest.py#L137

PR #519, opened 2024-10-10
Fix convert_timeslice function (take 2)

tsmbland:
# Description

this change.*

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/agents/__init__.py
+++ b/src/muse/agents/__init__.py
@@ -8,4 +8,4 @@ __all__ = [
 ]
 
 from muse.agents.agent import AbstractAgent, Agent, InvestingAgent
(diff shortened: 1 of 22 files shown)


tsmbland:
Thanks! Yeah I totally get your point about using a functional approach. I was less concerned about doing this because MUSE already deviates from this in many places, but that doesn't mean we shouldn't be doing things right when we can.

In fact, the global `TIMESLICE` object was always being used to begin with. The objects being passed around to reference their timeslice contained information about which timeslicing arrangement was required in the output, but information about timeslice length (which is required to split the quantity over the timeslices) has always come from `TIMESLICE` and not explicitly passed to the functions.

PR #533, opened 2024-10-24
Fix more broadcasting errors

tsmbland:
# Description

This PR fixes a few more bugs caused by automatic broadcasting.

I identified these using the approach suggested by @dalonsoa [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/518#pullrequestreview-2385012239), which works beautifully. It essentially turns off automatic broadcasting (I've limited this to the timeslice dimension), and throws an error message when broadcasting would be required to perform the intended operation. I've implemented this in a separate PR/branch (#530). I then ran the code in that branch and went through every place where it complained about automatic broadcasting, and was able to spot a few more places where `convert_timeslice` should be applied to split quantities over the timeslice dimension (rather than automatic broadcasting as was being done previously).

I then went back and applied most of these changes here. Long run we'll want to merge #530 to properly turn off automatic broadcasting (rather than just fixing errors like I'm doing here). However, there are lots of places where it ends up complaining about benign things, and it will still take some work to make it completely happy. Right now I just want to fix these bugs as quickly as possible. (#530 is also way ahead of this one and has a completely different `timeslices` module which isn't ready yet.)

Summary of the changes:

- Fix an error in the ALCOE function where fuel costs and env costs for the whole year were applied to each timeslice. This ends up effecting commodity prices
- Fix a similar error in the `emission_cost` objective
- Fix a similar error in the `emission` function, where emissions for the whole year were being applied to every timeslice. This mostly has implications for models using the carbon budget, as emissions will now be much lower than before. I've updated the carbon budget tutorial to reflect this
- Fix a similar error in `gross_margin` with the calculation of consumption and production costs

There's at least one more bug of this type remaining (in the `consumption` function), but I'm going to tackle that in a separate PR as it has much larger implications
Fixes # (issue)

## Type of change

## Key checklist

## Further checks


tsmbland:
Actually I'm mistaken about the `emission` function. It was correct before. Going to change it back.

Same thing with the error in `consumption` that I alluded to

tsmbland:
Closing this and opening #534 instead

PR #548, opened 2024-11-04
Delete old/redundant outputs

tsmbland:
Deletes outputs that aren't required any more, as discussed in #413 

Deleting these because they are known to be unreliable (e.g. #312), and all the relevant data can be created with other global/sector outputs

I advised in the release notes for v1.2.0 that these would eventually be deleted, so nobody should still be using these anyway.

Will save this for v1.3.0 along with #507 #510 and others


src/muse/outputs/mca.py
```diff
@@ -233,463 +233,6 @@ def _aggregate_sectors(
...
-@register_output_quantity(name=["timeslice_supply"])
-def metric_supply(
-    market: xr.Dataset, sectors: list[AbstractSector], **kwargs
-) -> pd.DataFrame:
-    """Current timeslice supply across all sectors."""
-    market_out = market.copy(deep=True)
-    return _aggregate_sectors(sectors, market_out, op=sector_supply)
-
-
-def sector_supply(sector: AbstractSector, market: xr.Dataset, **kwargs) -> pd.DataFrame:
```
dalonsoa, line 326: The infamous `sector_supply` function, gone for good!

src/muse/outputs/mca.py
```diff
@@ -233,463 +233,6 @@ def _aggregate_sectors(
...
-@register_output_quantity(name=["yearly_supply"])
-def metricy_supply(
-    market: xr.Dataset, sectors: list[AbstractSector], **kwargs
-) -> pd.DataFrame:
-    """Current yearlysupply across all sectors."""
-    market_out = market.copy(deep=True)
-    return _aggregate_sectors(sectors, market_out, op=sectory_supply)
-
-
-def sectory_supply(
```
dalonsoa, line 395: What the heck is `sectory_supply`? I hadn't seen this before...

dalonsoa (approved):
This changes all make sense, specially because they remove ambiguity. 

Remember to mention in the release note for the next release that these outputs are no longer available.

PR #550, opened 2024-11-04
Timeslice level

tsmbland:
# Description

this change.*

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/sectors/sector.py
+++ b/src/muse/sectors/sector.py
@@ -109,6 +109,7 @@ class Sector(AbstractSector):  # type: ignore
         from muse.production import maximum_production
 
         self.name: str = name
+        self.timeslice_level = ""
         """Name of the sector."""
         self.subsectors: Sequence[Subsector] = list(subsectors)
         """Subsectors controlled by this object."""
@@ -276,18 +277,25 @@ class Sector(AbstractSector):  # type: ignore
 
         # Calculate supply
         supply = self.supply_prod(
-            market=market, capacity=capacity, technologies=technologies
+            market=market,
+            capacity=capacity,
+            technologies=technologies,
+            timeslice_level=self.timeslice_level,
         )
 
         # Calculate consumption
-        consume = consumption(technologies, supply, market.prices)
+        consume = consumption(
+            technologies, supply, market.prices, timeslice_level=self.timeslice_level
+        )
 
         # Calculate commodity prices
         technodata = cast(xr.Dataset, broadcast_techs(technologies, supply))
         costs = supply_cost(
             supply.where(~is_pollutant(supply.comm_usage), 0),
             annual_levelized_cost_of_energy(
-                prices=market.prices.sel(region=supply.region), technologies=technodata
+                prices=market.prices.sel(region=supply.region),
+                technologies=technodata,
+                timeslice_level=self.timeslice_level,
             ),
             asset_dim="asset",
         )
--- a/src/muse/sectors/sector.py
+++ b/src/muse/sectors/sector.py
@@ -7,7 +7,6 @@ from typing import (
     cast,
 )
 
-import pandas as pd
 import xarray as xr
 
 from muse.agents import AbstractAgent
@@ -27,7 +26,6 @@ class Sector(AbstractSector):  # type: ignore
         from muse.outputs.sector import factory as ofactory
         from muse.production import factory as pfactory
         from muse.readers.toml import read_technodata
-        from muse.timeslices import TIMESLICE
         from muse.utilities import nametuple_to_dict
 
         # Read sector settings
@@ -38,9 +36,6 @@ class Sector(AbstractSector):  # type: ignore
             raise RuntimeError(f"Missing 'subsectors' section in sector {name}")
         if len(sector_settings["subsectors"]._asdict()) == 0:
             raise RuntimeError(f"Empty 'subsectors' section in sector {name}")
-
-        # Timeslices
-        timeslices = TIMESLICE.timeslice
 
         # Read technologies
         technologies = read_technodata(settings, name, settings.time_framework)
@@ -93,7 +88,6 @@ class Sector(AbstractSector):  # type: ignore
             name,
             technologies,
             subsectors=subsectors,
-            timeslices=timeslices,
             supply_prod=supply,
             outputs=outputs,
             interactions=interactions,
@@ -105,7 +99,6 @@ class Sector(AbstractSector):  # type: ignore
         name: str,
         technologies: xr.Dataset,
         subsectors: Sequence[Subsector] = [],
-        timeslices: pd.MultiIndex | None = None,
         interactions: Callable[[Sequence[AbstractAgent]], None] | None = None,
         interpolation: str = "linear",
         outputs: Callable | None = None,
@@ -121,10 +114,6 @@ class Sector(AbstractSector):  # type: ignore
         """Subsectors controlled by this object."""
         self.technologies: xr.Dataset = technologies
         """Parameters describing the sector's technologies."""
-        self.timeslices: pd.MultiIndex | None = timeslices
-        """Timeslice at which this sector operates.
-        If None, it will operate using the timeslice of the input market.
-        """
         self.interpolation: Mapping[str, Any] = {
             "method": interpolation,
             "kwargs": {"fill_value": "extrapolate"},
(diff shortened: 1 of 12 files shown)


tsmbland:
@dalonsoa Not finished yet, but just wondering why ruff is complaining about my type annotations? There are plenty of other places in the code using the `Optional[str] = None` syntax and it's not complaining about those. And I can't use `X | Y` like it's suggesting (even though plenty of places use this syntax as well without issues) because it then fails on Python 3.9

dalonsoa:
I believe I've found similar issues in the past. In principle, you can use `X | Y` annotation style even in Python 3.9 as long as you include `from __future__ import annotations` at the top of the relevant files. We do that in, for example, [PyCSVY](https://github.com/ImperialCollegeLondon/pycsvy/blob/develop/csvy/readers.py), where we support python 3.9 and use `X | Y` for annotations.

dalonsoa:
What you cannot have is the use of `__future__` and old style annotations. If you have it, you must use the new style. For example, [in `costs.py`](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/6171d32c7ab3e12a573f1d4dad9f69416a689927/src/muse/costs.py#L26), we don't have that import statement and you can perfectly use `Optional[str]`, but as you have the `future` import [in `constraints.py`](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/6171d32c7ab3e12a573f1d4dad9f69416a689927/src/muse/constraints.py#L99), that is not allowed. I think. That has always been very confusing.

src/muse/sectors/sector.py
```diff
@@ -103,25 +105,38 @@ def __init__(
...
+
         """Parameters describing the sector's technologies."""
+        self.technologies: xr.Dataset = technologies
+        if "timeslice" in self.technologies.dims:
+            if not get_level(self.technologies) == self.timeslice_level:
+                raise ValueError(
+                    f"Technodata for {self.name} sector does not match "
+                    "the specified timeslice level for that sector "
+                    f"({self.timeslice_level})"
+                )
```
dalonsoa, line 132: I agree with your implementation because it is explicit, but just in case, have you considered pulling the `timeslice_level` directly from the technodata rather than opening the possibility - that you check here - of having technodata timeslice information inconsistent with the declared timeslice level?

src/muse/sectors/sector.py
```diff
@@ -138,20 +153,26 @@ def __init__(
 
         :py:mod:`muse.interactions` contains MUSE's base interactions
         """
+        if interactions is None:
+            interactions = interaction_factory()
+        self.interactions = interactions
```
dalonsoa, line 158: Merely stylistic, so up to you, but I tend to use in this cases:
```suggestion
        self.interactions = interactions or interaction_factory()
```
But I'm not entirely sure if this works in older Python versions...

src/muse/sectors/sector.py
```diff
@@ -138,20 +153,26 @@ def __init__(
...
+
+        """Computes production as used to return the supply to the MCA.
 
         It can be anything registered with
         :py:func:`@register_production<muse.production.register_production>`.
         """
-        self.output_data: xr.Dataset
+        self.supply_prod = (
+            supply_prod if supply_prod is not None else maximum_production
+        )
```
dalonsoa, line 172: Same here:
```suggestion
        self.supply_prod = supply_prod or maximum_production
```

src/muse/sectors/sector.py
```diff
@@ -276,24 +302,51 @@ def market_variables(self, market: xr.Dataset, technologies: xr.Dataset) -> Any:
...
+            market["prices"], level=self.timeslice_level, operation="mean"
+        )
+        return xr.Dataset(dict(supply=supply, consumption=consumption, prices=prices))
+
+    def convert_to_global_timeslicing(self, market: xr.Dataset) -> xr.Dataset:
+        """Converts market data to global timeslicing."""
+        supply = expand_timeslice(market["supply"], operation="distribute")
+        consumption = expand_timeslice(market["consumption"], operation="distribute")
+        costs = expand_timeslice(market["costs"], operation="broadcast")
+        return xr.Dataset(dict(supply=supply, consumption=consumption, costs=costs))
```
dalonsoa, line 348: Very nice!

src/muse/timeslices.py
```diff
@@ -105,29 +76,191 @@ def setup_module(settings: Union[str, Mapping]):
...
-        if x.timeslice.reset_coords(drop=True).equals(ts.timeslice):
-            return x
+    if level is not None:
+        ts = compress_timeslice(ts, ts=ts, level=level, operation="sum")
+
+    # If x already has timeslices, check that it matches the reference timeslice.
+    if "timeslice" in data.dims:
+        if data.timeslice.reset_coords(drop=True).equals(ts.timeslice):
+            return data
         raise ValueError("x has incompatible timeslicing.")
```
dalonsoa, line 104: Now it is not `x` but `data`. Maybe also add a more descriptive message, eg.:
```suggestion
        raise ValueError("Data has incompatible timeslicing with target timeslice during broadcast operation.")
```
Or something along these lines.

src/muse/timeslices.py
```diff
@@ -105,29 +76,191 @@ def setup_module(settings: Union[str, Mapping]):
...
+        ts = TIMESLICE
+
+    if level is not None:
+        ts = compress_timeslice(ts, ts=ts, level=level, operation="sum")
+
+    # If x already has timeslices, check that it matches the reference timeslice.
+    if "timeslice" in data.dims:
+        if data.timeslice.reset_coords(drop=True).equals(ts.timeslice):
+            return data
+        raise ValueError("x has incompatible timeslicing.")
```
dalonsoa, line 136: As above.

src/muse/timeslices.py
```diff
@@ -105,29 +76,191 @@ def setup_module(settings: Union[str, Mapping]):
...
+
+    broadcasted = broadcast_timeslice(data, ts=ts)
+    timeslice_fractions = ts / broadcast_timeslice(ts.sum(), ts=ts)
+    return broadcasted * timeslice_fractions
```
dalonsoa, line 140: Just to make sure I get this right:
- First we broadcast the input array over the target timeslice - same value in all timeslices if it had no timeslice info, to start with
- Then, we multiply by the timeslice fractions to ensure that the data is proportional to the timeslice fraction.

src/muse/timeslices.py
```diff
@@ -105,29 +76,191 @@ def setup_module(settings: Union[str, Mapping]):
...
+
+    """
+    if ts is None:
+        ts = TIMESLICE
+
+    # Raise error if x is not timesliced appropriately
+    if "timeslice" not in data.dims:
+        raise ValueError("x must have a 'timeslice' dimension.")
+    if not data.timeslice.reset_coords(drop=True).equals(ts.timeslice):
+        raise ValueError("x has incompatible timeslicing.")
```
dalonsoa, line 170: As above. No `x` anymore.

dalonsoa (approved):
I've left a collection of small comments here and there, but I don't think any of them is a blocker, so I'm approving. 

This looks really neat and clear. Great work!

src/muse/timeslices.py
tsmbland: Correct

src/muse/sectors/sector.py
tsmbland: I agree this is way better

src/muse/sectors/sector.py
tsmbland: This would be possible for sectors where timeslice-level utilization factor and/or minimum service factor are specified. However, for sectors where this isn't specified (i.e. using a single UF/MSF for all timeslices), there's no timeslice information in the technodata so nothing to pull from, and these sectors can take any timeslice level.

In the end, it's easier just to be explicit about it for all sectors.

docs/inputs/toml.rst
```diff
@@ -319,6 +319,32 @@ A sector accepts these attributes:
    Additional methods can be registered with
    :py:func:`muse.production.register_production`
 
+*technodata*
```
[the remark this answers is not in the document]
tsmbland: I wish we were using markdown as it's so much nicer to work with, but I guess it's too late for that

Issue #475, opened 2024-09-06
[BUG] Unexpected behaviour in multi-region models


tsmbland:
## Describe the bug

I haven't done much work with multi-region models, but the example I've been working with ([tutorial 3](https://muse-os.readthedocs.io/en/documentation/user-guide/add-region.html)) doesn't look like I would expect.

Specifically, this model has two regions with gas, power and residential sectors. We're putting a strict upper limit on the capacity of heatpumps in region 2, and as a result, the residential sectors in the two regions look very different. I would therefore expect that, since the demand for gas and electricity will be different in the two regions, the gas and power sectors should have different capacities in the two regions to reflect this (I haven't added any commodity trade between regions). However, this isn't the case, and the gas and power sectors look identical in the two regions (at least according to the data in the output files). 

## To Reproduce

Run the model for tutorial 3 (found in `docs/tutorial-code/3-add-region`)

I don't know whether the problem is specific to this model or all multi-region models. If the latter, I'm surprised nothing has been flagged before.

## Expected behavior

I would expect the gas and power sectors in the two regions to have different capacities, reflecting the different demand for gas and electricity in the residential sector. Unless I'm completely misunderstanding something here...

## Screenshots

Follow the link above for tutorial 3 to see figures

## Context

Please, complete the following to better understand the system you are using to run MUSE.

- Operating system (eg. Windows 10):
- MUSE version (eg. 1.0.1): dev
- Installation method (eg. pipx, pip, development mode): dev
- Python version (you can get this running `python --version`):

tsmbland:
See https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/556#issuecomment-2550941689

PR #556, opened 2024-11-06
Fix commodity price calculations

### LINKED ISSUE #552: Commodity price calculations incorrect for technologies with output quantities not equal to 1
Commodity prices are calculated using the `annual_levelized_cost_of_energy` function, which calculates the cost of operating one unit of a technology at full capacity. However, it doesn't take output quantities into account, essentially assuming that one unit of capacity produces one unit of output. But this may be (very) incorrect. For a output quantity x, commodity prices will end up being x-fold higher than they should be.

Basically this function needs a complete rewrite to be more like the `lifetime_levelized_cost_of_energy` function:
- take capacity into account
- use exponents
- take output quantities into account

Also, see #551

### LINKED ISSUE #607: [BUG] Prices differ according to the timeslice level
See #550. I noticed that commodity prices change according to the timeslice level, in direct proportion to the number of timeslices. This should not be happening, so this is clearly highlighting an error somewhere in the calculation of commodity prices, although it's not totally clear where the problem lies.

In any case, this is a good control to make sure everything's working properly when the time comes.

tsmbland:
# Description

This PR aims to address a number of issues in the `costs` module to make the code cleaner and fix some incorrect calculations.

The main motivation for this was that models are currently particularly sensitive to the scaling of technology inputs and outputs, and often fail spectacularly if the output quantity isn't 1 (#552). In reality, we'd expect the scaling of inputs/outputs in the CommIn/CommOut files to have very little impact on the results, as long as inputs and outputs are scaled together - in other words it's the _ratio_ that matters and not the absolute values. I've added a test to prove that this is fixed (at least for the LCOE calculation).

I've also noticed that #607 to do with the timeslice level is fixed, which is a good sign

Detailed changes below. 

The results have now changed for all models, particularly commodity prices which now look a bit closer to what's expected as far as I can tell (electricity and gas prices more closely match the values in the projections file, which I assume are at least on the expected order of magnitude). I can't exactly pinpoint which changes are most responsible for changing the results as there's a lot going on. I'm actually a bit surprised that prices have changed this much as none of these models have technologies with output quantities not equal to 1. But I've double checked and can't see any mistakes in what I've done, so I have to assume it's all fine.

Note: We're currently still calculating costs at the timeslice level, so this will all change again. But it will be based on what I've done here, so this can still be reviewed and merged. 

## Main changes

Combining lifetime and annual LCOE into the same function 
- Capital costs are calculated for the lifetime of the technology (a one time cost), whereas running costs, production and revenue are calculated for a single year
- If the "annual" method is being used, capital costs are annualised by multiplying by the capital recovery factor
- If the "lifetime" option is being used, running costs and production are broadcasted across the lifetime of the technology, multiplied by the discount factor for each year, and then summed across years
- Annual LCOE is now based on actual capacity, production and consumption data, rather than effectively assuming that these were all equal to 1

Separate functions for cost components (e.g. fuel costs, environmental costs)
- Reduces code duplication in the costs module
- Can also be used in the objectives module to avoid duplicating the calculations here. 

Cost functions only take and return data for a single year
- Since #602, there's no need to pass data to these functions for a span of years. All costs are calculated based on data for a single year (usually the investment year).
- I'm therefore mandating that these functions are only passed data for a single year. This simplifies some aspects of the code (e.g. no interpolation), and avoids potentially unwanted broadcasting if some inputs have a year dimension and some don't
- I'm not actually checking whether the inputs have a year dimension, I'm just checking that the output doesn't have a year dimension (using a new decorator), which it would do if any arrays used in the calculation had a year dimension.
- There's no longer need for a `year` argument for any of the cost functions

Passing consumption data to cost functions rather than calculating it internally
- I think it gives greater separation of concerns if this data is calculated externally and passed to the function
- It also allows us to use _actual_ consumption data in some circumstances (especially the calculation of commodity prices) rather than always using made-up data
- This leads to slightly more repetitive code in the objectives module, as consumption is now calculated here, but I think it's worth it - and this could probably be simplified later

There was a `+ 1` applied to the year argument for the discount factor calculations ([here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/339ea03d9692f20fe5ded1338bccdfaf443e42f8/src/muse/costs.py#L272)), which meant that the discount factor was being applied in the first year of the technology lifetime. I think this is incorrect, so I've remove this `+ 1` (i.e. `year = 0` for the first year, so discount factor = 1)

Fixes #552 
Fixes #607 

## Type of change

## Key checklist

## Further checks


tsmbland:
I suspect there may be a mistake in the treatment of regions, as I've noticed that prices are abnormally high for the example models with multiple regions. I'm looking into it...

dalonsoa:
I'm holding the review until you have sorted the issue with the regions.

tsmbland:
@dalonsoa As far as I can tell it's not something I've done wrong in this PR, but the treatment of regions is generally a mess. It brings me back to issue #475

I think generally the problem is this:

In arrays with an "asset" dimension (e.g. supply, consumption), "region" should always be a property of the asset (not sure if that's the correct xarray term) rather than its own dimension. This is because each asset belongs to a specific region. However, the supply data passed to the LCOE function has "region" and "asset" as separate dimension, in which all the assets in each region have been copied across both regions. Ultimately LCOE is ending up approximately twice as high in the the model with two regions, which carries through to commodity prices. 

Basically, I don't think there's any good reason for any array to have both "asset" and "region" dimensions. I might try adding a check to the `patched_broadcast_compat_data` function and that should flag where the problems are throughout the code. But I don't know how easy it will be to fix.

I suspect the reason prices were previously similar in multi-region models was just a case of two bugs cancelling each other out. I suspect one of the bugs was the same one that was causing #607, which is now fixed, although I haven't fully figured out why.

In any case, I think you can go ahead and review this PR when you have the time.

tsmbland:
@dalonsoa #609 builds on this (and I think simplifies things overall), so it might be worth reviewing them both together

src/muse/costs.py
```diff
@@ -1,33 +1,245 @@
...
 
 
+def cost(func):
+    """Decorator to validate the output dimensions of the cost functions."""
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        result = func(*args, **kwargs)
+        assert "year" not in result.dims
+        assert "timeslice" in result.dims
```
dalonsoa, line 72: Following the description in the docstring above, maybe you could check that `commodity` is not a dimension, either.

src/muse/costs.py
dalonsoa, line 1: Assuming that the results are correct - which seem to be the case for what you say - this looks way, way better than it was before. Much clearer code, much clear execution pipeline, and much clearer explanations. Nothing to add, here, to be honest.

tests/test_costs.py
dalonsoa, line 1: I guess that coming up with tests that check some other properties of the outputs of the cost functions, besides the dimensions, is tricky, right?

dalonsoa (approved):
I've made a couple of comments/questions, but this looks really good. A much simplified and easy to follow code - and therefore, easier to figure out if things go wrong in the future.

tests/test_costs.py
tsmbland: Yeah it's a bit tricky. A lot of the test elsewhere just copy code directly from the respective function and check that the output is the same - which is completely pointless. I've added some tests for LCOE which demonstrate how LCOE is expected to change (or not) as the inputs are changed. I couldn't really think of similar tests for the other costs (or didn't try that hard), but since LCOE uses a lot of the individual cost components I think the coverage is still pretty good.

PR #557, opened 2024-11-08
Remove unused production methods

tsmbland:
As discussed with @ahawkes, removing these production methods as they're not being used and just getting in the way

Eventually I'm planning to use production values directly from the solver, but this is just step 1


tsmbland:
@dalonsoa Going to merge this without review as it's very simple, just keeping you in the loop

PR #560, opened 2024-11-11
:sparkles: Add log to file.

### LINKED ISSUE #498: Output warnings to a file

dalonsoa:
# Description

To make everyone's life easier when running and developing MUSE. Currently the tests also produce log files. This might or might not be desirable.

Fixes #498 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/inputs/toml.rst
+++ b/docs/inputs/toml.rst
@@ -66,9 +66,6 @@ a whole.
 *log_level*
    verbosity of the output. Valid options, from the highest to the lowest level of verbosity, are: "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL".
 
-*log_file*
-   Path to a file where the log will be saved, in addition to printing it to the console. If not given, the log will be added to a file named `muse_CURRENT_DATETIME.log` in the current working directory.
-
 *equilibirum_variable*
    whether equilibrium of `demand` or `prices` should be sought. Defaults to `demand`.
 
--- a/docs/inputs/toml.rst
+++ b/docs/inputs/toml.rst
@@ -66,6 +66,9 @@ a whole.
 *log_level*
    verbosity of the output. Valid options, from the highest to the lowest level of verbosity, are: "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL".
 
+*log_file*
+   Path to a file where the log will be saved, in addition to printing it to the console. If not given, the log will be added to a file named `muse_CURRENT_DATETIME.log` in the current working directory.
+
 *equilibirum_variable*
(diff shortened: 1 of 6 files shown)


tsmbland:
Thanks for this!

A few comments/ideas (happy to open these as separate issues if you don't want to do this here):
- Is it possible to have two separate files: one for INFO statements and one for WARNING statements? I especially want users to pay more attention to warnings, so I think having these in a separate file would help.
- I would just call these `muse_info.log` and `muse_warning.log` by default (i.e. no date). Otherwise if you run the model many times you're going to get an accumulation of log files (maybe this is what you'll want in some circumstances, but I'd say not by default) 
- I think it probably makes more sense to put these files in the results folder by default (i.e. `default_output_dir`). This way it's clear which log files match up with which results files, which could otherwise get confusing if you're running multiple models from the same working directory

dalonsoa:
No problem to do so now. And also to fix the tests, that I had forgotten that we support Python 3.9.

So, just to confirm, we don't give the option to select the log file name(s) just `muse_info` and `muse_warning`. I prefer that, simpler and cleaner, but just to be sure.

tsmbland:
> No problem to do so now. And also to fix the tests, that I had forgotten that we support Python 3.9.
> 
> So, just to confirm, we don't give the option to select the log file name(s) just `muse_info` and `muse_warning`. I prefer that, simpler and cleaner, but just to be sure.

Cool. Yeah let's just keep it simple and hardcode the file names

dalonsoa:
Done. I had to do some tweaking since tests were failing when running in parallel because they were all trying to access the same log file. I've disabled logging to file during tests, I think, since I could not find a way of indicating that using fixtures (all my attempts were ignored).

dalonsoa:
I've no idea why building the docs fail. Certainly, running `muse --model default` works locally (at least in Windows) so I'm not sure why it might be failing there.

tsmbland:
> I've no idea why building the docs fail. Certainly, running `muse --model default` works locally (at least in Windows) so I'm not sure why it might be failing there.

It's because the results folder doesn't exist yet when it tries to create the file (you'll already have a results folder locally from previous runs)

dalonsoa:
@tsmbland , all good, now! Many thanks for the tip.

tsmbland (approved):
Great, all looks good! Thanks very much

PR #563, opened 2024-11-12
Fix calculation of material costs

tsmbland:
# Description

this change.*

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/costs.py
+++ b/src/muse/costs.py
@@ -91,6 +91,9 @@ def net_present_value(
     products = is_enduse(technologies.comm_usage)
     fuels = is_fuel(technologies.comm_usage)
 
+    # Calculate consumption
+    cons = consumption(technologies=techs, production=production, prices=prices)
+
     # Revenue
     prices_non_env = filter_input(prices, commodity=products, year=years.values)
     raw_revenues = (production * prices_non_env * rates).sum(("commodity", "year"))
@@ -112,8 +115,7 @@ def net_present_value(
 
     # Fuel/energy costs
     prices_fuel = filter_input(prices, commodity=fuels, year=years.values)
-    fuel = consumption(technologies=techs, production=production, prices=prices)
-    fuel_costs = (fuel * prices_fuel * rates).sum(("commodity", "year"))
+    fuel_costs = (cons * prices_fuel * rates).sum(("commodity", "year"))
 
     # Cost related to material other than fuel/energy and environmentals
     prices_material = filter_input(prices, commodity=material, year=years.values)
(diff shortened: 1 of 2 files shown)


src/muse/costs.py
```diff
@@ -112,34 +115,39 @@ def net_present_value(
...
+        convert_timeslice(
+            techs.fix_par * (capacity**techs.fix_exp),
+            prices.timeslice,
+            QuantityType.EXTENSIVE,
+        )
+        * rates
+    ).sum("year")
+
+    # Variable costs
+    tech_activity = (production / techs.fixed_outputs).max("commodity")
```
dalonsoa, line 135: Here we are no longer selecting the `products`, as before. I guess this is correct?

src/muse/costs.py
```diff
@@ -276,31 +287,45 @@ def lifetime_levelized_cost_of_energy(
...
     )
-    variable_costs = (
-        techs.var_par * production.sel(commodity=products) ** techs.var_exp
-    ).sum("commodity")
-    fixed_and_variable_costs = ((fixed_costs + variable_costs) * rates).sum("year")
-    denominator = production.where(production > 0.0, 1e-6)
+
+    # Production
+    prod = (
+        production.where(production > 0.0, 1e-6)
```
dalonsoa, line 314: I always thought this `1e-6` was quite arbitrary. Not sure what a better way of dealing with it would be, though.

dalonsoa (approved):
The changes look sensible to me. 

Having said that, I don't get why changing to use consumption instead of production - which I agree was wrong - does not change a single number in any of the models. Surely, this should be affecting the results, right?

src/muse/costs.py
tsmbland: Previously it was adding together the production of all `products`, then using this to scale the variable costs (as if the variable costs should be applied per unit of produced commodity, which is wrong). 

I'm doing something different which is calculating a ratio between commodity production and the output per unit of activity (`fixed_outputs`), and using this to scale the variable costs. In theory this ratio should be the same for all output commodities, whether `products` or `environmentals` (and nan for all commodities that aren't outputted), so when I take the max it should be fine without explicitly selecting `products`, but I guess there's no harm in doing so and it would probably be safer/more readable.

tsmbland:
> Having said that, I don't get why changing to use consumption instead of production - which I agree was wrong - does not change a single number in any of the models. Surely, this should be affecting the results, right?

This change only affects the calculation of material costs, which is for any commodity flagged by `is_material`. The short answer is that none of the commodities in any of the models are flagged as material. I'm pretty sure it's all to do with the `CommodityType` specified in the `GlobalCommodities.csv` file, but I'd have to double check how this works.

PR #580, opened 2024-11-27
Enforce consistency in objective dimensions

### LINKED ISSUE #581: [BUG] Odd results for objectives that don't return an `asset` dimension
Some of the objectives (namely "capital_costs", "comfort", "efficiency" and "ALCOE") don't return an `asset` dimension, and I've noticed some strange behavior when using these objectives (e.g. not investing in the technologies you'd expect), or more often the model hits a growth error and fails to run.

I think the place where this is causing problems is the `_unified_dataset` function, although that function is so unreadable it's hard to see what it's supposed to be doing.

Simply adding an `asset` dimension to these objectives by broadcasting looks like it's enough to fix the problem (#580), and gives much more sensible results. That may end up being the best solution. 

I've also noticed that this bug only applies when using these objectives as single objectives (or multi-objective with each other). I guess that when you combine these with another objective that does have an `asset` dimension, then automatic broadcasting occurs at that point, so by the time it get's passed to `_unified_dataset` it's all good.

tsmbland:
# Description

See #581 for a description of the problem.

This PR enforces that all objectives must have an `asset` dimension. I've added these checks in the decorator, and modified the "comfort", "efficiency", "capital_costs" and "ALCOE" objectives accordingly, adding an `asset` dimension by broadcasting. The tests also needed updating.

Now, all objectives are consistent in the dimensions they return, apart from some of them which also return a `timeslice` dimension. However, this ends up being compressed by this line:

https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/106d7acde6d4578643b2e2d076d6dc7b5c6e166f/src/muse/investments.py#L312

So now ultimately by the time the objectives reach the solver the dimensions are always the same: `asset` and `replacement`.

This seems to fix the problems described in the issue (see case study below), although I haven't really figured out why. I think it's to do with the scipy adapter (`_unified_dataset` and `lp_constraint` and `lp_constraint_matrix` functions), although the code is so unreadable it's hard to figure out what's going on. I'm not sure it matters though for now, as long as the issue is fixed.

To help enforce dimensions, I've created the `check_dimensions` helper function. This is checking both the inputs and outputs of the objectives. I think this is something that could be used in several other places throughout the code, but I'll look into this separately.

I've suppressed two tests that were failing. I haven't fully understood why these are failing, but they're both do do with noninvesting agents which will no longer calculate objectives in v1.3 (#554), so I'm not too fussed about it.
 
### Case study

I've taken the default model and changed the objective to "capital_costs". I've also increased the capital costs of `gasboiler` 10x to 30, so the expected outcome is clear - it should invest in heatpumps in the residential sector

#### Scenario 1: original code, single objective (capital costs)

The solver fails to find a solution. This is strange as changing the objective is only supposed to change the costs vector, which shouldn't effect whether or not a solution is possible (this is all determined by the constraints)

#### Scenario 2: original code, adding a dummy second objective (LCOE, 0.01 weight)

Now the model runs to completion and the results are in line with expectations. 

Capacity_Original_MultiObjective

I think this works because when the objectives are combined, the `asset` dimension from LCOE gets picked up in the overall combined objective.

#### Scenario 3: new code, single objective (capital costs)

Now we get the expected result, even without adding a dummy objective

Capacity_New_Capital

Fixes #581 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.3
+current_version = 1.2.4rc1
 commit = True
 tag = True
 
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.2.4rc1
+current_version = 1.2.3
 commit = True
 tag = True
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.7.4
+    rev: v0.8.0
     hooks:
       - id: ruff-format
         types_or: [python, pyi, jupyter]
@@ -17,7 +17,7 @@ repos:
         types_or: [python, pyi, jupyter]
         args: [--fix]
   - repo: https://github.com/igorshubovych/markdownlint-cli
-    rev: v0.42.0
+    rev: v0.43.0
     hooks:
       - id: markdownlint-fix
   - repo: https://github.com/kynan/nbstripout
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -9,5 +9,5 @@ authors:
       given-names: Adam
 
 title: MUSE_OS
-version: v1.2.3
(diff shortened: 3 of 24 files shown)


tsmbland:
Actually don't review this, I've opened up #584 instead which is pretty much the same but based on the v1.3 branch which means the tests no longer fail

PR #584, opened 2024-11-28
Enforce consistency in objective dimensions

### LINKED ISSUE #581: [BUG] Odd results for objectives that don't return an `asset` dimension

tsmbland:
# Description

See #581 for a description of the problem.

This PR enforces that all objectives must have an `asset` dimension. I've added these checks in the decorator, and modified the "comfort", "efficiency", "capital_costs" and "ALCOE" objectives accordingly, adding an `asset` dimension by broadcasting. The tests also needed updating.

Now, all objectives are consistent in the dimensions they return, apart from some of them which also return a `timeslice` dimension. However, this ends up being compressed by this line:

https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/f94ee1bdc0c96f5c7a211f970766fb73f7ff6192/src/muse/investments.py#L289

So ultimately by the time the objectives reach the solver the dimensions are always the same: `asset` and `replacement`.

This seems to fix the problems described in the issue (see case study below), although I haven't really figured out why. I think it's to do with the scipy adapter (`_unified_dataset` and `lp_constraint` and `lp_constraint_matrix` functions), although the code is so unreadable it's hard to figure out what's going on. I'm not sure it matters though for now, as long as the issue is fixed.

To help enforce dimensions, I've created the `check_dimensions` helper function. This is checking both the inputs and outputs of the objective functions. I've also used this in the `demand_share` module, and there are probably many other places it could be used, but I'll leave it like this for now.
 
### Case study

I've taken the default model and changed the objective to "capital_costs". I've also increased the capital costs of `gasboiler` 10x, so the expected outcome is clear - it should invest in heatpumps in the residential sector.

#### Scenario 1: original code, single objective (capital costs)

The solver fails to find a solution. This is strange as changing the objective is only supposed to change the costs vector, which shouldn't effect whether or not a solution is possible (this is all determined by the constraints)

#### Scenario 2: original code, adding a dummy second objective (LCOE, 0.01 weight)

Now the model runs to completion and the results are in line with expectations. 

Capacity_Original_MultiObjective

I think this works because when the objectives are combined, the `asset` dimension from LCOE gets picked up in the overall combined objective.

#### Scenario 3: new code, single objective (capital costs)

Now we get the expected result, even without adding a dummy objective

Capacity_New_Capital

Fixes #581 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -311,9 +311,9 @@ def test_check_dimensions():
     check_dimensions(data, required=["dim1"], optional=["dim2"])
 
     # Missing required
-    with raises(ValueError):
+    with raises(ValueError, match="Missing required dimensions"):
         check_dimensions(data, required=["dim1", "dim3"], optional=["dim2"])
 
     # Extra dimension
-    with raises(ValueError):
+    with raises(ValueError, match="Extra dimensions"):
         check_dimensions(data, required=["dim1"])
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -1,6 +1,6 @@
 import numpy as np
 import xarray as xr
-from pytest import approx, mark
+from pytest import approx, mark, raises
 
 
 def make_array(array):
@@ -296,3 +296,24 @@ def test_avoid_repetitions():
     assert 3 * len(result.year) == 2 * len(assets.year)
     original = result.interp(year=assets.year, method="linear")
     assert (original == assets).all()
+
+
+def test_check_dimensions():
+    from muse.utilities import check_dimensions
+
+    data = xr.DataArray(
+        np.random.rand(4, 5),
+        dims=["dim1", "dim2"],
+        coords={"dim1": range(4), "dim2": range(5)},
+    )
+
+    # Valid
+    check_dimensions(data, required=["dim1"], optional=["dim2"])
+
+    # Missing required
+    with raises(ValueError):
+        check_dimensions(data, required=["dim1", "dim3"], optional=["dim2"])
+
+    # Extra dimension
+    with raises(ValueError):
+        check_dimensions(data, required=["dim1"])
--- a/src/muse/demand_share.py
+++ b/src/muse/demand_share.py
@@ -120,7 +120,9 @@ def factory(
         result = function(agents, market, technologies, **keyword_args)
 
         # Check result
-        check_dimensions(result, ["asset", "timeslice", "commodity"])
+        check_dimensions(
+            result, ["timeslice", "commodity"], optional=["asset", "region"]
+        )  # asset should be required
         return result
 
     return cast(DEMAND_SHARE_SIGNATURE, demand_share)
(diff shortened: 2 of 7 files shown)


tests/test_utilities.py
```diff
@@ -296,3 +296,24 @@ def test_avoid_repetitions():
...
+        np.random.rand(4, 5),
+        dims=["dim1", "dim2"],
+        coords={"dim1": range(4), "dim2": range(5)},
+    )
+
+    # Valid
+    check_dimensions(data, required=["dim1"], optional=["dim2"])
+
+    # Missing required
+    with raises(ValueError):
```
dalonsoa, line 314: Check the error text as well to make sure it fails for the right reason:
```suggestion
    with raises(ValueError, match="Missing required dimensions"):
```

tests/test_utilities.py
```diff
@@ -296,3 +296,24 @@ def test_avoid_repetitions():
...
+        check_dimensions(data, required=["dim1", "dim3"], optional=["dim2"])
+
+    # Extra dimension
+    with raises(ValueError):
```
dalonsoa, line 318: Same here:
```suggestion
    with raises(ValueError, match="Extra dimensions"):
```

dalonsoa (approved):
LGTM! It seems like a good runtime verification with minimal overhead.

tsmbland:
@dalonsoa One thing I'm a bit unsure about is when it's better to use `assert` statements and when it's better to explicitly raise an error, like I'm doing here. There are lots of places in the code using `assert` to check dimensions, which seems like it's effectively the same, but is there an important distinction?

dalonsoa:
There's nothing wrong with `assert`, but raising an error let you control better the type of error (so it is more informative to the user).

It also happens that assertions are ignored when [running python in optimised mode](https://docs.python.org/3/tutorial/modules.html#compiled-python-files) (with the -O and -OO flags). I've never known anyone using these flags for anything, so I'm not sure if it is a big deal, but you should not trust the behaviour of your code to a feature that might be disabled by the users. In summary, assert should be used only in tests or for debugging purposes.

PR #586, opened 2024-11-29
Fix inconsistent reading of agent share names

tsmbland:
# Description

Agent shares are specified in the agents files, and then referenced in technodata files to assign a share of existing capacity to the agent. The names specified in the technodata files must match the agents file, otherwise an error will be raised. However, as @Jin-Jun-Mei found, in some cases even when the names do match an error can still be raised. 

The reason is that the process for reading in the agent share names is different for the agents file and technodata files. For the agents file (`read_csv_agent_parameters`), names are converted using a `sub` operation. For technodata files (`read_technodictionary`), names are first converted to snake case and then converted using a similar `sub` operation. If following conventional naming schemes, these end up being equivalent, but if you deviate from this you can end up with inconsistencies and MUSE complaining that you haven't specified an agent share which you have.

I've simplified this so that in both cases were just converting to snake case. This should leave no room for inconsistency, as long as the names match in the files. I've updated the documentation to make it clear that names should match between the files.

Also the trade model needed updating as the names didn't match between the files

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/inputs/technodata.rst
+++ b/docs/inputs/technodata.rst
@@ -136,32 +136,31 @@ InterestRate
    is the technology interest rate (called hurdle rates in other models).
    This is used for the interest used in the discount rate and corresponds to the interest built when borrowing money.
 
-Agent_0, ..., Agent_N
-   represent the allocation of the initial capacity to the each agent.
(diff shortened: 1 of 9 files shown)


dalonsoa (approved):
Nice and simple.

PR #378, opened 2024-06-27
Changes default solver to scipy

dalonsoa:
# Description

Sets the default solver to scipy, updating also the settings files.

Fixes # N/A

## Type of change

Please add a line in the relevant section of
[CHANGELOG.md](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/development/CHANGELOG.md) to
document the change (include PR #) - note reverse order of PR #s.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (non-breaking, back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (whatever its nature)

## Key checklist

- [x] All tests pass: `$ python -m pytest`
- [x] The documentation builds and looks OK: `$ python -m sphinx -b html docs docs/build`

## Further checks

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #587, opened 2024-11-29
Change default solver to scipy

tsmbland:
# Description

Revisits the changes is #378 to change the default solver to scipy.

I think now is a good time to do this since v1.3 already has lots of changes that will require people to properly read the release notes and make changes to their settings files. If anyone misses this, that's their fault.

I've also added a warning if `lpsolver` is not specified in their settings file, to tell people that it's defaulting to scipy.

I had to change one of the tests as it was failing with the scipy solver. Not sure why (or why this wasn't a problem in #378), but switching to adhoc for this one test allows it to run.


docs/inputs/toml.rst
```diff
@@ -442,12 +442,12 @@ Sectors contain a number of subsections:
         are registered via :py:func:`~muse.investments.register_investment`. At time of
         writing, three are available:
 
+        - "scipy" solver (default): Formulates investment as a true LP problem and solves it using
```
dalonsoa, line 445: Maybe indicate from when it is the default solver?
```suggestion
        - "scipy" solver (default from v1.3): Formulates investment as a true LP problem and solves it using
```

docs/inputs/toml.rst
```diff
@@ -442,12 +442,12 @@ Sectors contain a number of subsections:
...
+          the `scipy solver`_.
+
         - an "adhoc" solver: Simple in-house solver that ranks the technologies
           according to cost and service the demand incrementally.
 
-        - "scipy" solver: Formulates investment as a true LP problem and solves it using
-          the `scipy solver`_.
-
         - "cvxopt" solver: Formulates investment as a true LP problem and solves it
```
dalonsoa, line 451: Is this actually working? I've never tried it or had to deal with it.

dalonsoa (approved):
LGTM!

docs/inputs/toml.rst
tsmbland: A quick test suggests the answer is no (at least in the current MUSE version - it's not covered by the tests so might have been accidentally broken along the way)

PR #591, opened 2024-12-03
Add release notes to documentation

### LINKED ISSUE #538: Add release notes to documentation
We should have a page in the documentation to store the release notes for all the different releases. Release notes can currently be found on github (e.g. [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/releases/tag/v1.2.1)), but it would be nice to have everything in to documentation as well. We can then get into the habit of writing release notes as part of pull requests

tsmbland:
# Description

Adds release notes for versions 1.2.0 to 1.2.3 to the documentation. No problems using markdown, so could pretty much copy and paste these from github. It's all completely manual, but I think that's fine as these should be tailored to users rather than automatically built from PR names or commit messages

I also found a way to link to github issues/PRs without writing out the full URL

You can see what this all looks like [here](https://muse-os.readthedocs.io/en/documentation/)

Fixes #538

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -69,6 +69,20 @@ html_theme = "classic"
 
 
 def github_role(name, rawtext, text, lineno, inliner, options=None, content=None):
+    """Creates links to issues/pull requests on the MUSE_OS GitHub site.
+
+    To use in markdown:
+    {github}`ISSUE_NUMBER` (e.g. {github}`123`)
+
+    To use in rst:
+    :github:`ISSUE_NUMBER` (e.g. :github:`123`)
+
+    In both cases this will create a clickable link (visible as #123) to the relevant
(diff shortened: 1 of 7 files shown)


docs/conf.py
```diff
@@ -61,3 +64,29 @@
 # -- Options for HTML output -------------------------------------------------
 
 html_theme = "classic"
+
+# -- Render GitHub links -------------------------------------------------
+
+
+def github_role(name, rawtext, text, lineno, inliner, options=None, content=None):
```
dalonsoa, line 71: Brilliant!!

dalonsoa (approved):
Really nice! And the documentation page looks very professional. 

Fancy writing a small blogpost on creating useful release notes and including them in the docs 😄 ?

PR #595, opened 2024-12-05
Perform the utilisation factor check on the whole technodata dataset

### LINKED ISSUE #513: Utilization factor column mandatory in `Technodata.csv` even when specified in `TechnodataTimeslice.csv`
If UF is specified in `TechnodataTimeslice.csv`, any UF specified in the `Technodata.csv` file will be overwritten. Nevertheless, MUSE will complain if this column is missing in the `Technodata.csv` file

dalonsoa:
# Description

After all relevant files have been loaded, as per @tsmbland suggestion. I spotted what felt like a left-over R3 in the trade technodata and that cause the tests to fail. 

The approach is not the most elegant one - take thw whole xarray dataset, transform it to a pandas dataframe and perform the checks as before. Not the best performance-wise, but as this is done only once at the begining, performance is not an issue so opted for the path of minimum effort. 

Fixes #513

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/readers/csv.py
+++ b/src/muse/readers/csv.py
@@ -92,8 +92,6 @@ def read_technodictionary(filename: Union[str, Path]) -> xr.Dataset:
     data = data.drop(["process_name", "region_name", "time"], axis=1)
     data = data.apply(to_numeric, axis=0)
 
-    check_utilization_and_minimum_service_factors(data, filename)
-
     result = xr.Dataset.from_dataframe(data.sort_index())
     if "fuel" in result.variables:
         result["fuel"] = result.fuel.isel(region=0, year=0)
@@ -140,7 +138,6 @@ def read_technodata_timeslices(filename: Union[str, Path]) -> xr.Dataset:
     data = csv[csv.technology != "Unit"]
 
     data = data.apply(to_numeric)
-    check_utilization_and_minimum_service_factors(data, filename)
 
     ts = pd.MultiIndex.from_frame(
         data.drop(
@@ -402,6 +399,10 @@ def read_technologies(
     result = result.set_coords("comm_usage")
     if "comm_type" in result.data_vars or "comm_type" in result.coords:
         result = result.drop_vars("comm_type")
+
+    check_utilization_and_minimum_service_factors(
+        result.to_dataframe(), [tpath, ttpath]
+    )
 
     return result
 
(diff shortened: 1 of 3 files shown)


src/muse/readers/csv.py
```diff
@@ -869,7 +870,7 @@ def check_utilization_and_minimum_service_factors(data, filename):
     if "utilization_factor" not in data.columns:
         raise ValueError(
             f"""A technology needs to have a utilization factor defined for every
-             timeslice. Please check file {filename}."""
+             timeslice. Please check files: {filename}."""
```
tsmbland, line 873: A few of the other functions further down need a similar change

src/muse/readers/csv.py
```diff
@@ -403,6 +400,10 @@ def read_technologies(
     if "comm_type" in result.data_vars or "comm_type" in result.coords:
         result = result.drop_vars("comm_type")
 
+    check_utilization_and_minimum_service_factors(
+        result.to_dataframe(), [tpath, ttpath]
```
tsmbland, line 404: Ideally we should only include `ttpath` is it's not `None`, otherwise we get `None` printed as a path in the error message

tsmbland (changes requested):
Thanks! I think this is definitely the easiest solution and it does the job. Not too fussed about performance here as, like you say, it only happens once.

Just a couple of very minor changes requested

src/muse/readers/csv.py
```diff
@@ -403,6 +400,10 @@ def read_technologies(
     if "comm_type" in result.data_vars or "comm_type" in result.coords:
         result = result.drop_vars("comm_type")
 
+    check_utilization_and_minimum_service_factors(
```
AdrianDAlessandro, line 403: Just a note for if performance is a concern: Since this function is only ever used here, the internals of it could be adjusted to take an Xarray instead of a DataFrame

dalonsoa: It is not an issue because it's used only once and the size of the arrays is pretty small, so the overhead is minimal. I've given a try to adapt the code to use xarrays and it has become less readable - and didn't work (granted I only invested 15 min).

In summary, I'd leave it as it is.

src/muse/readers/csv.py
dalonsoa: sorted

src/muse/readers/csv.py
dalonsoa: sorted

tsmbland (approved):
Happy with this, thanks!

PR #596, opened 2024-12-05
Remove bumpversion

### LINKED ISSUE #435: Remove use of `bump2version`
Currently we're using `bump2version` whenever we want to bump the version number in order to make a new release. This process is also currently not documented #436.

I've just had a fairly frustrating experience with @tsmbland trying to make it create a new `rc` release, without success. Thinking about it, I'm not sure we need `bump2version` at all. Nowadays you can just specify the version number in `pyproject.toml`; if you need to access the version number from the source code you can get it with `importlib` anyway.

AdrianDAlessandro:
# Description

This PR removes `bump2version` from the repo. It also makes a small change to enable dynamically getting the version from the git tag instead of needing to specify it in the `__init__.py`. This is done using [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/usage/).

This has been applied to everywhere the version is required except for the `CITATION.cff`, which can maybe be updated automatically by including a script, but since both the version and the release date are required, it might not be able to be automated. There could be a way to check the versions are consistent in the release workflow.

I also rename from `VERSION` to `__version__` as I believe this is more standard for version strings.

The the steps for making a new release should only be:
1. Update the `version` and `date-released` fields in `CITATION.cff`
2. Make a release on GitHub with a tag matching the version number in the citation file

This is evidence of it working:
```bash
$ python -c "import muse; print(muse.VERSION)"
1.3.1.dev5+g436f2b4a
$ git tag 2.2.2
$ pip install -e .
...
Successfully built MUSE_OS
Installing collected packages: MUSE_OS
  Attempting uninstall: MUSE_OS
    Found existing installation: MUSE_OS 1.3.1.dev5+g436f2b4a
    Uninstalling MUSE_OS-1.3.1.dev5+g436f2b4a:
      Successfully uninstalled MUSE_OS-1.3.1.dev5+g436f2b4a
Successfully installed MUSE_OS-2.2.2
$ python -c "import muse; print(muse.VERSION)"
2.2.2
```

Fixes #435 

## Type of change

## Key checklist

## Further checks


src/muse/__init__.py
```diff
@@ -1,8 +1,12 @@
 """MUSE model."""
 
 import os
+from importlib.metadata import PackageNotFoundError, version
 
-VERSION = "1.3.0"
+try:
+    __version__ = version("MUSE_OS")
+except PackageNotFoundError:
+    pass
```
dalonsoa, line 9: As mentioned in the python template repo, the following would be best if we are to ignore the exception:
```suggestion
from contextlib import suppress

with suppress(PackageNotFoundError):
    __version__ = version("MUSE_OS")
```
Or giving it a default value.

dalonsoa (approved):
LGTM, but would be good checking if when building the standalone executables with `pyinstaller` (both, the command line tool and the GUI), the `__version__` value is setup properly. I think it would, but we should double check.

tsmbland:
Just tried running `python -c "import muse; print(muse.__version__)"` in this branch and I get `1.2.2rc1`

I guess I don't know what I'd expect since there's no tag for this branch, but strange that it's giving an old version...

tsmbland:
Ahhh I see I think `1.2.2rc1` was the last time I ran `pip install -e .` I understand now

AdrianDAlessandro:
> LGTM, but would be good checking if when building the standalone executables with `pyinstaller` (both, the command line tool and the GUI), the `__version__` value is setup properly. I think it would, but we should double check.

This is a good idea. I'm not sure how to build those executables, nor how to check the version after they're made. Can either of you check it? @tsmbland @dalonsoa

tsmbland:
> > LGTM, but would be good checking if when building the standalone executables with `pyinstaller` (both, the command line tool and the GUI), the `__version__` value is setup properly. I think it would, but we should double check.
> 
> This is a good idea. I'm not sure how to build those executables, nor how to check the version after they're made. Can either of you check it? @tsmbland @dalonsoa

To be honest I'm not quite sure either, but I'll look into it

tsmbland:
@dalonsoa  I've just realised we haven't been creating executables for any version since v1.1.0. The workflow is running https://github.com/EnergySystemsModellingLab/MUSE_OS/actions/runs/12139463002, and the artifact is created, but doesn't appear as an asset on the release page https://github.com/EnergySystemsModellingLab/MUSE_OS/releases

Any idea what's going on?

tsmbland:
I've built the executables locally and it seems fine getting the version - at least for the GUI the expected version is showing, I'm not actually sure how to get the version number from the command line executable, but I assume it's fine

dalonsoa:
> @dalonsoa I've just realised we haven't been creating executables for any version since v1.1.0. The workflow is running https://github.com/EnergySystemsModellingLab/MUSE_OS/actions/runs/12139463002, and the artifact is created, but doesn't appear as an asset on the release page https://github.com/EnergySystemsModellingLab/MUSE_OS/releases
> 
> Any idea what's going on?

I've no idea. The code looks ok and, as you say, the artifact is uploaded properly, so I don't know why it is not pushed to the release page. I'll check what might be going on...

tsmbland (commented):
This is great. The only concern I have is that if someone's done a developer installation (i.e. cloning from github and running `pip install -e .`), then `__version__` will persist from when they ran the installation, and won't update with the code if they run `git pull` (at least that appears to be the case). Maybe not such a big issue as most people are installing from pypi, but it would be relevant for #597 which I've just opened. Curious if anyone has any ideas.

AdrianDAlessandro:
> This is great. The only concern I have is that if someone's done a developer installation (i.e. cloning from github and running `pip install -e .`), then `__version__` will persist from when they ran the installation, and won't update with the code if they run `git pull` (at least that appears to be the case). Maybe not such a big issue as most people are installing from pypi, but it would be relevant for #597 which I've just opened. Curious if anyone has any ideas.

Unfortunately, I believe that's a downside of using `importlib.metadata` to get the version. But if someone is working on the development version of the code, even if it says the version number, they may have edited the code anyway

tsmbland (approved):

PR #601, opened 2024-12-13
Fix the `consumption` function

tsmbland:
# Description

Broken this off from #556 as it's relatively simple and self contained

The `consumption` function takes production data and then back-calculates commodity consumption based on the technology `CommIn`. But the calculation was all wrong

Let's say you've got a technology with the following reaction: 1A -> 2B + 3C

If you present production data of 4B & 6C to `consumption`, you should get 2A (this is the amount of A you'd need to produce 4B and 6C). To calculate this, you first need to calculate the degree of technology activity (i.e. how many units of flow do you need to produce 4B and 6C), and then multiply this by the `CommIn` quantity (in this case 1).

But previously it was summing together the output quantities (assuming they are both end-use commodities) and multiplying this by `CommIn`, so it would have given 10A ((4 + 6) * 1)

The logic for correctly calculating the degree of technology activity is included in a new function `production_amplitude`. This same calculation is done a few times in the `costs` module, although I haven't used the function here to avoid conflicts with #556 

Hopefully `consumption` is now correct and more readable. This will have implications for commodity prices as fuel costs ans material costs depend on the amount of commodity consumption. It doesn't change any of the results in the example models as all the technologies have a single end-use output commodity.

### DIFF
--- a/src/muse/quantities.py
+++ b/src/muse/quantities.py
@@ -579,7 +579,7 @@ def production_amplitude(
         DataArray with production amplitudes for each technology in each timeslice.
         Will have the same dimensions as `production`, minus the `commodity` dimension.
     """
-    assert set(technologies.dims).issubset(set(production.dims))
+    # assert set(technologies.dims).issubset(set(production.dims))
 
     return (
         production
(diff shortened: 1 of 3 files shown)


dalonsoa (approved):
Looks good to me! The maths certainly make sense. It will be good, in the future, to have an example that actually uses this to illustrate how it works. I might even suggest you put part of your PR description in the docs - or the docstring - since it very well explains what the function does and how it does it.

PR #602, opened 2024-12-13
Limit prices used for objectives to investment year

tsmbland:
# Description

See first two points in [this comment](https://github.com/EnergySystemsModellingLab/MUSE_OS/discussions/466#discussioncomment-10487127). We want objectives to use the investment year commodity prices, and only these prices, when calculating current and future costs.

It's not obvious unless you enter debug mode, but `prices` has a "year" dimension where the first year is the current year and the second year is the investment year (previously the third year would have been the forecast year, if applicable, but I've accidentally broken this, see #603). I'm not entirely happy about selecting based on the index, but it's the best I can do for now without larger refactoring.

This is prep work for #556, where all the costs functions have been changed to only accept prices for a single year (which will be broadcast into the future - no interpolation)

@ahawkes Can you confirm that this is correct. i.e. when calculating costs over the lifetime of a technology (e.g. LCOE), we want to always assume flat-forward prices from the investment year - correct?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.2
+    rev: v0.8.3
     hooks:
       - id: ruff-format
         types_or: [python, pyi, jupyter]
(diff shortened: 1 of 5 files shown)


dalonsoa (approved):
Well, it is a very simple change, that makes the results correct, so even if not ideal, it seems cost effective.

ahawkes:
Yes this seems correct. (Well, not correct, but consistent). A better approach would be to have investment in the current year, rather than in the next period. But I think that change could be too complicated and is perhaps better left to MUSE 2.0.

PR #605, opened 2024-12-16
Split fixed costs in proportion to production

tsmbland:
# Description

I tried the suggestion [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/246#issue-2158533253) about splitting non-variable costs across timeslices in proportion to the output in each timeslice, but the result in the end is that commodity prices are now always equal in every timeslice. This makes sense because all costs are now proportional to production, so LCOE (= costs / production) will always be the same regardless of production. (I guess this would be different if any of the exponents in the costs calculations were not equal to 1, but this isn't the case in any of the example models).

@ahawkes Were you expecting something different here?
@dalonsoa 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


tsmbland:
Adding to this, even if LCOE is equal in every timeslice, prices can still be different in different timeslices as they're a weighted average of LCOE across technologies. I can see this in the minimum service factor tutorial, as the MSF forces different technology utilization in different timeslices.

In any case, I think it doesn't make any sense to calculate LCOE at the timeslice level (either for objectives or for calculating commodity prices). In this case, I think we should change all the calculations so that LCOE is only calculated at the annual level. I'll give this a go this unless I get any objection.

dalonsoa:
No objection form my side. I think LCOE only makes sense at annual level, but I seem to remember @ahawkes have reasons for needing it at a more granular level.

tsmbland:
This was the original comment about needing LCOE at the timeslice level: https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/246#issuecomment-2074651759

However, I tried the suggestion about changing the LCOE calculation, and ended up with identical LCOE in every timeslice. In which case there's no point calculating LCOE at the timeslice level if it's always going to be the same in every timeslice

PR #609, opened 2024-12-18
Calculate LCOE at the annual level

tsmbland:
# Description

Changed the costs module so that LCOE (and NPV, EAC etc.) are calculated at the annual level rather than the timeslice level. Some of the individual cost components that make up LCOE (such as fuel costs, environmental costs) are still calculated at the timeslice level, but these are summed over timeslices for the LCOE calculation.

The `patched_broadcast_compat_data` patch is very helpful and reassuring here, as it makes sure that no automatic broadcasting occurs over the timeslice dimension. This means that, even if we tried to add a non-timesliced cost (e.g. capital cost) to a timesliced cost (e.g. fuel cost) it would complain rather than automatically broadcasting the full capital cost over every timeslice.

Commodity prices are now the same in every timeslice, as expected, except for some models where the mix of technologies supplying the commodity is different in different timeslices (see electricity in the minimum service tutorial)

Builds on from #556


dalonsoa (approved):
Code wise, the changes look ok and, once again, result in simpler code, which is always good. I assume that the numeral result still make perfect sense - or hopefully more - right?

tsmbland:
I'm having second thoughts about this. Let's say that electricity is more expensive in one timeslice because demand is high so costlier technologies have to be used. If, for example, heat relies on electricity, then we'd also want the price of heat to go up in these timeslices to reflect the increasing electricity prices. But calculating the LCOE at the annual level there's no way to capture this.

So actually I think the solution in #605 is better - still calculating LCOE at the timeslice level, but splitting non-variable costs in proportion to the output in each timeslice.

I'm going to revert this PR, and we can revisit this in January.

PR #633, opened 2025-01-17
LCOE: split capital/fixed/variable costs over timeslices in proportion to production

### LINKED ISSUE #288: Change logic of how LCOE is used to calculate commodity prices

### LINKED ISSUE #613: Rethink the calculation of commodity prices
I'm having second thoughts about this. Let's say that electricity is more expensive in one timeslice because demand is high so costlier technologies have to be used. If, for example, heat relies on electricity, then we'd also want the price of heat to go up in these timeslices to reflect the increasing electricity prices. But calculating the LCOE at the annual level there's no way to capture this.

So actually I think the solution in #605 is better - still calculating LCOE at the timeslice level, but splitting non-variable costs in proportion to the output in each timeslice. Alternatively, we can explore other methods for calculating commodity prices (e.g. marginal costs)

I'm going to revert this PR, and we can revisit this in January.

_Originally posted by @tsmbland in https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/609#issuecomment-2554799760_

tsmbland:
# Description

This PR follows on from #609 (which I reverted), and is more along the lines of #605 (which I initially rejected, but now favor)

The main issues are whether LCOE should be calculated at the timeslice-level or for the year as a whole, and ensuring this calculation is correct. When calculating LCOE for objectives, it makes more sense to have a single LCOE value for the year which depends on annual utilization. This is addressed in a separate PR (#635), which should be reviewed after this one. 

However, it still makes sense to calculate LCOE at the timeslice-level when calculating commodity prices (see #613). LCOE is currently being calculated at the timeslice-level (and only ever at the timeslice-level), but the approach for doing this is wrong, which is what I'm trying to fix in this PR.

In particular, capital costs are being distributed evenly across timeslices (in proportion to timeslice-length), which artificially inflates LCOE (and therefore commodity prices) in timeslices with low output (as the capital cost per unit of output is higher). Realistically, the capital cost per unit of output should be the same in every timeslice, which means splitting costs in proportion to production rather than timeslice-length. The same is true for fixed costs and variable costs.

I've added a test which highlights the issue (`test_lcoe_equal_prices`). If commodity prices are equal in every timeslice, then LCOE should also be equal in every timeslice. This was not the case previously, due to the issue about per-unit capital cost described above.

Another outcome is that if production is zero in a timeslice, then all components of LCOE (fuel costs, capital costs etc.) will now be zero in that timeslice. In this case, LCOE is strictly undefined, but will end up being zero in the calculation due to [this line](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/f1d54cafab23b925d7a9c099e0e7928eb80e65ca/src/muse/costs.py#L442). This is captured in `test_lcoe_zero_production`. I think it's ok the LCOE gets set to zero, as timeslices with zero production will be excluded from the weighted average when calculating commodity prices. (And also fine for the objectives due to the `timeslice_max` operation [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/f1d54cafab23b925d7a9c099e0e7928eb80e65ca/src/muse/investments.py#L285), but this is changing anyway in #635 so I'll address this there). 

Smaller changes:
- I actually don't think it's possible to have nans or infs in the LCOE, so I've removed [this line](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/9146768ee33d89d4dfb018ba443d7f618175d72d/src/muse/objectives.py#L490), but added [this check](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/f1d54cafab23b925d7a9c099e0e7928eb80e65ca/src/muse/costs.py#L72) to be on the safe side.
- no longer passing `timeslice_level` to functions in the `costs` module. Most functions don't need this, and those that do can infer the timeslice level from passed inputs
- I've modified the `timeslices` module to support multi-dimensional timeslice weights (the `ts` argument to functions). This was necessary for splitting capital costs over timeslices, as different technologies require different timeslice weights.

Another important thing to remember is that operations between timesliced and non-timesliced arrays are prohibited. This precludes many potential timeslice-related calculation errors, such as accidentally applying the full annual capital cost to every timeslice.

Fixes #288 (partially, see also #635)
Fixes #613 

## Type of change

## Key checklist

## Further checks


src/muse/quantities.py
```diff
@@ -509,7 +507,6 @@ def capacity_to_service_demand(
 def production_amplitude(
     production: xr.DataArray,
     technologies: xr.Dataset,
-    timeslice_level: str | None = None,
```
dalonsoa, line 512: Need to update the docstring as well.

src/muse/timeslices.py
```diff
@@ -141,14 +143,17 @@ def distribute_timeslice(
 
     # If data already has timeslices, check that it matches the reference timeslice.
     if "timeslice" in data.dims:
-        if data.timeslice.reset_coords(drop=True).equals(ts.timeslice):
+        if data.timeslice.reset_coords(drop=True).equals(
+            ts.timeslice.reset_coords(drop=True)
+        ):
```
dalonsoa, line 148: Why was this change needed?

src/muse/costs.py
```diff
@@ -212,11 +201,24 @@ def running_costs(
...
-    _variable_costs = variable_costs(technologies, production, timeslice_level)
+
+    # Costs associated with capacity and production level (annual)
+    _fixed_costs = fixed_costs(technologies, capacity)
+    _variable_costs = variable_costs(technologies, production)
+
+    # Split fixed/variable across timeslices in proportion to production
+    timeslice_level = get_level(production)
+    tech_activity = production_amplitude(production, technologies)
+    _fixed_costs = distribute_timeslice(
```
dalonsoa, line 216: This is an unexpected use of `distribute_timeslice`, I think.

According to its docstring, the `ts` parameter should be a `Dataarray with timeslice lengths.`, but here you are giving `tech_activity`. I understand that you are using this as weights to distribute the costs, which makes sense, so maybe the problem is in the description of `distribute_timeslice`, which should clarify that `ts` are the weights assigned to each timeslice and not necessarily their lengths (in time units).

src/muse/costs.py
```diff
@@ -264,14 +265,17 @@ def net_present_value(
...
         xr.DataArray with the NPV calculated for the relevant technologies
     """
     # Capital costs (lifetime)
-    _capital_costs = capital_costs(
-        technologies, capacity, timeslice_level, method="lifetime"
+    _capital_costs = capital_costs(technologies, capacity, method="lifetime")
+
+    # Distribute capital costs across timeslices in proportion to production
+    tech_activity = production_amplitude(production, technologies)
+    _capital_costs = distribute_timeslice(
```
dalonsoa, line 277: So the whole weighting by the tech activity also applies to the NPV, right?

dalonsoa (approved):
I've left a couple of comments but, I think, they mostly boil down to docstrings, so approving. It has taken a couple of iterations, but I think the explanation makes sense.

src/muse/costs.py
tsmbland: Good point - ts refers to weights, not necessarily lengths. I've updated the docstring

src/muse/timeslices.py
tsmbland: I had to modify this function (and others) to support timeslice arrays with multiple dimensions. This is so that different timeslice weights can be used for different technologies (important when splitting costs based on production). No exactly sure why `reset_coords` is required for this, but it was giving errors without this. There were a couple of other changes required like doing `sum("timeslice")` to sum the timeslice weights instead of just `sum()`.

I've now modified the `timeslice` fixture in the tests so that both scenarios are tested - single dimensional and multi-dimensional timeslice arrays - all seems fine

PR #635, opened 2025-01-20
Aggregate LCOE and NPV over timeslices for objectives

### LINKED ISSUE #288: Change logic of how LCOE is used to calculate commodity prices

tsmbland:
# Description

Previously, LCOE was calculated at the timeslice-level for the objectives, and then a [max operation](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/9146768ee33d89d4dfb018ba443d7f618175d72d/src/muse/investments.py#L285) was taken over the timeslices, and this was the value taken as the cost of installing one unit of capacity.

It seems nonsensical to take the max, and could prevent a technology from being invested in just because it's expensive in a single timeslice.

Really, we should be calculating the average LCOE across the whole year and using this as the cost metric.

To do this, I've added an `aggregate_timeslices` argument to the LCOE function, which sums timeslice-level costs and production when `True`, returning a single value for the year (no "timeslice" dimension). This is then what's used for the objectives.

Ultimately, this then gets broadcast over the timeslices [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/9146768ee33d89d4dfb018ba443d7f618175d72d/src/muse/objectives.py#L149), and the same max operation is taken, which gives you back a single value proportional to the original LCOE for the year as a whole. 

It's all a bit convoluted, but this seemed like the easiest way to introduce these changes quickly. I think there are bigger changes required to the solver in the future which will require larger refactoring, so there's no point over-optimizing things right now.

Surprisingly, none of the results change at all. Don't know what to make of this, but it still seems like the right thing to do

Fixes #288 

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Well, it is a bit convoluted, but I don't feel it is a big bottleneck. The key thing is a question of documentation - possibly a separate PR: when LCOE is timeslice-dependent, when it is not and why that is the case.

tsmbland:
I've added similar changes to NPV/NPC/EAC

PR #638, opened 2025-01-23
Unpin xarray

### LINKED ISSUE #625: xarray pinned to <=2024.11
Was having problems with later versions of xarray, so have pinned for now. Ideally we should figure out the problem so we can use the latest versions

tsmbland:
For some reason, with new version of xarray, calling `.interp` on `technologies` causes the `commodities` dimension to be dropped as an indexable dimension, which means later on the program complains when trying to select technologies by commodity. Not sure if this is a bug or feature in xarray? Anyway, I've added a single line of code to fix this, which means xarray can now be unpinned

Fixes #625


dalonsoa (approved):
That's odd, but so be it. Maybe ask in the xarray repo if this is a feature or a bug.

PR #642, opened 2025-01-24
Tidy up the use of interpolation

### LINKED ISSUE #265: `interpolation_mode` parameter needs clarification (9.2)
The example settings file in 9.2 gives `interpolation_mode = 'Active'` as the default, but the meaning of this isn't explained. Instead, the documentation for `interpolation_mode` further down the page gives a load of other options (linear, nearest, cubic etc.), but no "Active" option. 

- What is the meaning of  `interpolation_mode = 'Active'`? What interpolation type is used in this case?
- Other parts of the documentation say: "Interpolation is activated only if the feature *interpolation_mode = 'Active'* is defined in the TOML file.". In which case, what happens if you use another option such as linear or cubic? Are these redundant/legacy?

tsmbland:
# Description

I've always been a bit bugged by the over-reliance on temporal interpolation. After all, there's a well defined time framework, which most objects should follow, so interpolation should rarely be required. Yet the code is riddled with unnecessary uses of `interp`.

I've found that most of these `interp` calls can be changed to `sel` without problem (in the case that the objects in question _do_ cover the appropriate years), whereas others uses of `interp` are still necessary (i.e. where the object doesn't cover the appropriate years).

I think it's far better to be explicit about this, as using `interp` where it's not necessary could hide genuine bugs.

I've also got rid of `filter_input`, since, without interpolation, this is equivalent to `sel` (although I'm keeping `AbstractAgent.filter_input` for now as this serves another purpose)

Interpolation is still required for the following:
- reading the initial market. The method is configurable with the `interpolation_mode` parameter. I've updated the documentation for this parameter
- technologies parameters. This previously had no method specified, although I believe it should have been using the `interpolation` method specified in the settings file. It now is and I've updated the documentation
- sometimes interpolation is still required for capacity datasets, as the time framework of these is based on the lifetime of assets which may not match the time framework - we should always use linear interpolation here, as explained in `cliff_retirement_profile`, so there's no need for sectors and agents to have an `interpolation` attribute. I've created the helper function `interpolate_capacity` to make this consistent
- some interpolation remains in the `regressions` module - I'm not really sure what this module is doing and don't want to change it, but I think it's fine for now

Not sure why, but this appears to have fixed a bug with prices in the trade model (at least I think it was a bug - R2 prices were suspiciously high before). Won't think too much about it now as that model still needs a lot of work, but encouraging

Fixes #265 

## Type of change

## Key checklist

## Further checks


src/muse/utilities.py
```diff
@@ -437,6 +402,25 @@ def avoid_repetitions(data: xr.DataArray, dim: str = "year") -> xr.DataArray:
     return data.year[years]
 
 
+def interpolate_capacity(
```
dalonsoa, line 405: Conceptually, I'm not entirely sure if a linear interpolation of the capacity makes much sense. Capacity should only change when there's an investment in the relevant technology, so it should look very much staircase, going up only in the investment years. Wouldn't `previous` make more sense?

dalonsoa (approved):
I've a question, but the changes make total sense to me. Performance wise, it might even be faster as, presumably, a lookup of the value(s) will be faster than interpolating. In any case, it is more explicit, for sure.

src/muse/utilities.py
tsmbland: You're completely right, yes. If the capacity profile was created with `cliff_retirement_profile`, which I believe most are, then there will be no difference, but in general forward filling makes more sense than linear interpolation.

I've created an issue about this (#656) and will return to it later. It's an easy change to make, but some of the results are affected and I'd like to understand why before merging.

PR #644, opened 2025-01-27
Remove clean_assets

tsmbland:
This doesn't have any impact apart from the trade model, which is broken anyway. It also uses the `agent.forecast_year` attribute which I'm trying to get rid of in #645, so it's just easier to delete the whole thing


src/muse/agents/factories.py
```diff
@@ -275,7 +275,7 @@ def _shared_capacity(
 
 
 def _standardize_inputs(
-    housekeeping: str | Mapping | Callable = "clean",
+    housekeeping: str | Mapping | Callable = "noop",
```
dalonsoa, line 278: Ehmmm?

dalonsoa (approved):
This looks OK, but just for clarity, what was this supposed to do and why it was not used? I see the docstrings of the removed functions. Is this cleaning not needed to run MUSE?

tsmbland:
"noop" means no operation.

By default, newcapa agents (used for most simulations) were using "noop", whereas retrofit and noninvesting agents were using "clean" (I don't know why it was done like this).

So the deleted code was only ever seen by some simulations

The purpose of the "clean" seems three-fold:
- interpolate capacity data so the required years are covered
- delete data from before the current year
- delete any assets that have zero capacity now and in the future

The first isn't required as this interpolation happens repeatedly elsewhere. The latter two aren't strictly necessary as this is just dropping unused data, but this could have an impact on performance, otherwise `self.assets` just ends up accumulating stale assets.

Thinking about this again, it might be better if we don't delete this, start using it for all models, and I can modify it so it just drops stale data without doing the interpolation (and this won't need the `forecast_year` attribute, which I'm trying to get rid of)

See #652

tsmbland:
Dropped in favour of #652

PR #645, opened 2025-01-27
Remove remains of the forecast feature

tsmbland:
This was originally used to allow agents to look beyond the investment year, but it wasn't working properly and I've had permission from Adam to remove this feature to help simplify the code. Most of the work was done in #631, but some remnants still remained which I couldn't get rid of until #657 as `agent.forecast_year` was being used in the outputs module.


dalonsoa (approved):
LGTM. I see there are no changes to any of the output files, so I guess `forecast` was totally useless, right? Was `forecast` the same as `foresight`? Also, when we talk about `investment year`, is that the current simulation year or the next one?

tsmbland:
> LGTM. I see there are no changes to any of the output files, so I guess `forecast` was totally useless, right? 

`forecast` was used in the past, but not any more.

> Was `forecast` the same as `foresight`? 

I think "forecast" and "foresight" mean the same thing (i.e. allowing agents to look forward into the future), although as long as I've been working on MUSE the `foresight` parameter hasn't been doing anything (apart from adding an extra year to the time framework).

I'm guessing maybe `forecast` replaced `foresight` at some point (`foresight` was a global parameter, whereas `forecast` was sector-level to allow finer control), and then `foresight` was never properly removed?

> Also, when we talk about `investment year`, is that the current simulation year or the next one?

The investment year is the next one on from the current year. E.g. agents in 2020 will make investments for 2025 based on information (demand, prices, technology parameters) they have about 2025

I actually think "current year" is a bit misleading as MUSE is really simulating a time window each iteration (i.e. 2020 to 2025), and has fully solved the system for that whole time window by the end of the iteration (not just the first/"current" year), so maybe "base year" or "planning year" would be clearer

In MUSE2 this is simpler - only the current year, where agents plan and make investments in the same year

PR #649, opened 2025-01-29
Add default_adhoc example model and regression test

tsmbland:
I think it's worth having a regression test using the adhoc solver, as some people are still using this so it's useful to know when the results change


dalonsoa (approved):
LGTM! Not sure if there's a section about it in the docs about the example models and what they are about, but if there is, it would need to be updated, as well.

tsmbland:
Thanks @dalonsoa 

I've managed to ignore the specific warning for the test, so all good to merge now

Good point about docs - there's a readme file in `src/muse/data`, which I've now added to, although there's nothing in the actual documentation. I've added an issue about this (#654)

PR #650, opened 2025-01-30
Further improvements to `broadcast_techs`

tsmbland:
# Description

This makes a number of changes to the `broadcast_techs` function and its usage, as a follow up to #585 

- Renamed the function to `broadcast_over_assets`, which better describes its general usage

- Clarified the docstring, added example for using `installed_as_year=False`

- Removed calls to this function away from several functions in the `quantities` module (`emission`, `consumption`, `maximum_production`,`capacity_in_use`, `minimum_production`). It is now the responsibility of the code calling these functions to make sure the data is in the correct format. However, due to the check introduced in #647, any errors due to inconsistent formatting (i.e. trying to combine asset-level data with technology-level data) will be flagged, so I don't think this is any less robust. Overall, I think this makes these functions more readable, flexible (so long as the inputs are in a consistent format) and reduces the number of times conversions have to happen, as a converted technologies can be passed to multiple functions.

- The above change required some reformatting to other parts of the code, particularly the `demand_share` module

- Removed `filter_with_template`. This always just called `broadcast_techs` (the code beneath the first return statement was never accessed), so isn't required.

- Updated the tests in line with these changes

There are still a couple of TODOs in `broadcast_over_assets`, but I'm leaving these for now as they have knock on consequences for other parts of the code (particularly the adhoc solver and constraints), which I'm not yet comfortable changing.

### DIFF
--- a/src/muse/__main__.py
+++ b/src/muse/__main__.py
@@ -114,8 +114,8 @@ def patched_broadcast_compat_data(self, other):
             "(e.g. a capacity dataset with an 'asset' dimension) with a fully explicit "
             "technology dataset (e.g. a technology dataset with 'region' and "
             "'technology' dimensions). "
-            "Please use `broadcast_techs` on the latter object before performing this "
-            "operation."
+            "Please use `broadcast_over_assets` on the latter object before performing "
+            "this operation."
         )
 
     return self_data, other_data, dims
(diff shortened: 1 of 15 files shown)


dalonsoa (approved):
LGTM. The code is more readable with the use of `technologies` instead of a bunch of random words to define the same thing and the purpose of the broadcast function clearer 👍

PR #657, opened 2025-02-06
Save outputs for investment year

### LINKED ISSUE #540: Wrong year being used for LCOE and EAC outputs
The LCOE and EAC functions have a `year` parameter which determines which year to calculate the metric for. When calling these functions to generate the output files, `agent.year` is passed as the `year` parameter. However, this is incorrect, as by this point `agent.year` has already been incremented to the next year in the time framework. Thus, the LCOE/EAC values in the output files actually correspond to the wrong year.

tsmbland:
Following on from #641, in this PR we're saving data for the _investment year_ each iteration, rather than the _current year_. The exception is the first iteration, where we save output data for both years. This way, we get output data for every year of the time framework, without having to simulate an extra year (as was done before #641). As far as the user is concerned nothing has changed compared to before #641 - they're still getting the same data for all the same years, but it will be slightly quicker without the extra time step.

It's still a bit messy and wasteful. Most outputs are calculated for both years, but only saving data for the investment year. For now I think it's fine, but certainly more work still to be done.

I've made some changes to `outputs.mca` so the functions take a `year` argument, and some changes dotted through the code to pass around this year argument. Due to the way that the `save_multiple_outputs` is reused multiple times I also had to add `**kwargs` to lots of functions taht don't use `year`, so they don't complain about being passed `year` against their will. Not very satisfying, but the simplest fix without having to refactor a load of stuff.

Just noticed the cached outputs feature. I've made some changes to keep the tests happy, but it seems like it was broken anyway to begin with (only outputting data for the first year). Not really sure why this exists anyway.

This also fixes #540 which was caused by inconsistency in the year used for various parts of the LCOE/EAC calculations - now fixed as we're passing `year` to the functions and using this throughout.

Close #540


src/muse/outputs/cache.py
dalonsoa, line 1: If I remember correctly, the purpose of this was to save intermediate results of the simulation for post-mortem analysis. I believe I was the one setting it up :P , and I think back then it was working fine but, as many things in MUSE, it might have broken along the way and, if no one uses the feature (and there are no thorough tests), no one notice it is broken.

src/muse/outputs/sector.py
```diff
@@ -104,10 +104,10 @@ def _factory(
 
     def save_multiple_outputs(market, *args, year: int | None = None) -> list[Any]:
         if year is None:
-            year = int(market.year.min())
+            year = int(market.year.max())
```
dalonsoa, line 107: This seems like a radical change, from min to max. Why is it? Is it because `min` is current year and `max` is investment year or something like that?

dalonsoa (approved):
There are a lot of years and kwargs floating around, but all in all, the changes seem reasonable. At least the MCA is more explicit on the year we are talking about.

src/muse/outputs/sector.py
tsmbland: Yes, exactly. This is actually never used in the main code as I've made sure `year` is always passed explicitly. There are just a couple of tests that rely on this, but they weren't super straightforward to change so I've left this in (for now)

PR #658, opened 2025-02-07
Deepcopy market object to prevent side-effects in outputs

tsmbland:
I've noticed that some of the outputs have side-effects because they modify the `market` object. They are currently performing a shallow copy on the market object before modifying, but it turns out a deep copy is required to prevent side effects 

I just noticed this because I tried adding the LCOE output to a model and found that suddenly data was missing from some of the other outputs in 2025.


tsmbland:
@dalonsoa Going to merge this without review, but an interesting (and slightly scary) observation

dalonsoa:
It is indeed scary!!!

PR #668, opened 2025-03-11
Make capacity growth parameters optional

### LINKED ISSUE #667: Make `MaxCapacityAddition`, `MaxCapacityGrowth` and `TotalCapacityLimit` optional
I've often seen users setting these to very high values (e.g. 100000000) so that these limits effectively don't apply. Better would be to make these columns optional and just don't apply the constraint if the column is omitted. Even better would be to allow "inf" as a value so that the constraint can be applied on a per-technology basis, but this might be a bit harder to implement.

tsmbland:
# Description

This is a small change to make the MaxCapacityAddition, MaxCapacityGrowth and TotalCapacityLimit parameters optional.

I've found that many users just set these to very high values so that the constraint effectively doesn't apply, but this is annoying and messy.

It's already possible to exclude this constraint by modifying `constraints` in settings.toml, but this is a bit fiddly and in general I don't think users should be modifying `constraints` (I encourage users to leave this out of their settings files so all the constraints are used by default).

I think the change here to make these parameters optional is more user friendly, as all users need to do now is delete the relevant columns if they don't want to use the constraint for that sector.

Close #667 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/inputs/technodata.rst
+++ b/docs/inputs/technodata.rst
@@ -81,7 +81,7 @@ var_par, var_exp
 
    ProductionRef is the production of a reference capacity (CapRef) for the cost estimate decided by the modeller before filling the input data files.
 
-Growith constraints
+Growith constraints (optional)
    MaxCapacityAddition
       represents the maximum addition of installed capacity per technology, per year in a period, per region.
 
--- a/docs/inputs/technodata.rst
+++ b/docs/inputs/technodata.rst
@@ -107,6 +107,8 @@ Growith constraints (optional)
    Growth constraints are applied for each single agent in a multi-agent simulation. When only one agent is present, the growth constraints
    apply individually to the "New" and "Retrofit" agent, when present.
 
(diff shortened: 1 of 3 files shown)


tsmbland:
> Assuming the solver is happy with `np.inf`, and it seems it is considering that the tests run, this looks good to me.

The solver won't accept `np.inf`, however (unless the user puts "inf" in the input file, see below), `np.inf` should never reach the solver
- if all three parameters are not specified, the constraint isn't constructed (returns `None`)
- if one (or two) of the parameters is missing, this will be interpreted as `np.inf`, but should always be replaced by a finite value in the line `b = np.minimum(np.minimum(add_cap, total_cap), growth_cap)`	

I only just realised while making this PR that "inf" is an accepted value for these parameters, and it will work as intended unless all three parameters are "inf" (then the min operation will return `np.inf` and it will fail). This is why I haven't mentioned in the documentation that "inf" can be used, so hopefully users won't do this.

PR #670, opened 2025-03-11
Fix `same_enduse` and `spend_limit` filters

tsmbland:
Tiny fix, but the `same_enduse` and `spend_limit` filters were failing because they tried to filter the `technologies` dataset by year, but it's already been filtered by year [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/3a0bb4aebcaaba28d41dd3d3b8f9806ed01d1b96/src/muse/sectors/sector.py#L204) so complains when trying to use these filters.

I've fixed that bug. 

Also added a check to the decorator to enforce that `technologies` cannot have a year dimension. This was a bit trickier than usual as not all filters take `technologies`, so I had to get creative using `inspect`. Not sure if there's a better way to do this.

Had to make a few other changes to keep the tests happy

EDIT: Switched to the approach suggested by Diego below

### DIFF
--- a/src/muse/filters.py
+++ b/src/muse/filters.py
@@ -90,6 +90,7 @@ __all__ = [
     "with_asset_technology",
 ]
 
+import inspect
 from collections.abc import Mapping, MutableMapping, Sequence
 from typing import (
     Any,
@@ -134,6 +135,14 @@ def register_filter(function: SSF_SIGNATURE) -> Callable:
     def decorated(
         agent: Agent, search_space: xr.DataArray, *args, **kwargs
     ) -> xr.DataArray:
+        # Check inputs
+        sig = inspect.signature(function)
+        params = sig.parameters
+        if "technologies" in params:
+            bound_args = sig.bind(agent, search_space, *args, **kwargs)
+            technologies = bound_args.arguments["technologies"]
+            assert "year" not in technologies.dims
+
         result = function(agent, search_space, *args, **kwargs)  # type: ignore
         if isinstance(result, xr.DataArray):
             result.name = search_space.name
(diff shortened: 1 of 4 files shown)


src/muse/filters.py
```diff
@@ -134,6 +135,13 @@ def register_filter(function: SSF_SIGNATURE) -> Callable:
     def decorated(
```
dalonsoa, line 135: This looks fine and should work, so all good from my side. 

An alternative approach, possibly more explicit, would be to force for arguments other than `agent` and `search_space` to be passed as keyword arguments. It would look something like:

```python
    def decorated(
        agent: Agent, search_space: xr.DataArray, **kwargs
    ) -> xr.DataArray:
        # Check inputs
        if "technologies" in kwargs:
            assert "year" not in kwargs["technologies"].dims

        result = function(agent, search_space, **kwargs)  # type: ignore
        if isinstance(result, xr.DataArray):
            result.name = search_space.name
        return result
    return decorated
```
You might want to change the actual filters - and the calling code - to ensure that all arguments but for the first two are passed as keyword arguments, like:

```python
@register_filter
def same_enduse(
    agent: Agent,
    search_space: xr.DataArray,
    *,  # <-- This is the key
    technologies: xr.Dataset,  # <-- This now needs to be passed as keyword arg
    **kwargs,
) -> xr.DataArray:
    ...
```

dalonsoa (approved):
This looks good. I've made a suggestion that I think might make the code simpler and more explicit, but up to you if you want to take it. Your implementation looks fine as well.

src/muse/filters.py
tsmbland: This is neat! I didn't know you could do this...

PR #671, opened 2025-03-12
Error message for inf growth constraints

tsmbland:
See discussion in #668


dalonsoa (approved):
LGTM!

dalonsoa:
Aside from the failing tests, which I've no idea why is happening.

dalonsoa:
It might be worth at some point to create an autoused fixture that sets the seed of the random number generator to a reasonable number, because this randomness is a pain (it always has been).

Issue #371, opened 2024-06-25
Odd results when existing capacity data doesn't match the time framework [BUG]


tsmbland:
One of the tutorials talks about changing the time framework to every 2 years instead of every 5 years. According to the tutorial, this is a simple matter of changing the settings file, and doesn't involve changing any of the timing data in the csv files. Even though the existing capacity is specified every 5 years, MUSE should carry out linear interpolation and everything should still be fine with a 2-year framework, right?

Annoyingly this doesn't seem to be the case. If I switch the time framework to every two years without changing the capacity files, I get a very strange result in the gas sector:

![capacity_base](https://github.com/EnergySystemsModellingLab/MUSE_OS/assets/23723407/28e7cd47-f9a4-4d29-a925-0553b719f05c)
![capacity_base_2years](https://github.com/EnergySystemsModellingLab/MUSE_OS/assets/23723407/22030197-15f8-4447-b08a-76b68e875f01)

If I interpolate on the capacity files myself (to create a new set of files matching the new time framework), and then re-run the model, it looks much more reasonable:

![capacity_base_2years_interp](https://github.com/EnergySystemsModellingLab/MUSE_OS/assets/23723407/93f978ad-17f4-49f0-a53f-8a1566e0ca89)

I have no idea what the root cause of the issue is, but it seems like something is going wrong when the capacity data doesn't match up with the time framework (even though MUSE _should_ deal with this fine by interpolating the capacity data)

A couple of approaches:
- Enforce that the timing data in the capacity files must match the time framework specified in the settings file (and update the tutorial model accordingly)
- Find out what's causing the bug in this example and fix it

ahawkes:
Yes your interpretation is right.

I favour the latter solution - find out what is causing the weird result and fix it.

PR #672, opened 2025-03-13
Restructure input data for example models

### LINKED ISSUE #666: Simplify folder structure of input data
I think the folder structure of the input data is overly complicated, with too many nested folders. Better would be to change the `input` folder to `global_inputs`, which is more descriptive, and have a top-level folder for each sector (no `technodata` folder). All of this is configurable in the settings file, and there's currently nothing stopping users from doing it this way, but I think we should change the example models to make things easier for new users

tsmbland:
# Description

Restructuring the example models to be clearer for new users.

Easiest just to look at the new version of the [default model](https://github.com/EnergySystemsModellingLab/MUSE_OS/tree/cd41f9c69e85d160e24470a3d20be5000c4569ef/src/muse/data/example/default)

Main changes:
- got rid of the "Technodata" folder - sector folders now live at the top level
- renamed the "preset" folder to "residential_presets" to match the sector name
- got rid of the "inputs" folder - these files now live at the top level (along with the "Agents.csv" file in cases where agents are shared between sectors)
- removed the "BaseYearImport" and "BaseYearExport" files as these weren't being used
- removed some optional parameters from the settings file (lpsolver, constraints, dispatch_production, demand_share)
- restructured the settings file to a more logical order (e.g. outputs at the end)
- updated `wizard.py` to work with the new structure
- deleted `test_factory_constraints_passed_to_agents` because it was annoying me (oops)

This is just about changing the example models rather than the code, so shouldn't have any effect on other models. Users can still structure their files however they want as long as they link their files properly in the settings file, but I think the structure here is simpler/less convoluted than what we had before so should be clearer for new users.

Notes:
- for some very strange reason this changes the results for the `modify-time-framework` tutorial model. I can't conceive why, but that model was already giving strange results (see #371) so I'm not too concerned, and perhaps there might be a clue here for fixing that issue

Fixes #666


tsmbland:
@dalonsoa This one's probably annoying to review and I'm keen to avoid marge conflicts so going to merge now

PR #675, opened 2025-03-14
Set random seed for tests

tsmbland:
# Description

We sometimes find that tests randomly fail due to issues with randomly generated test data. Ideally we'd design the tests/fixtures in a way that wasn't so sensitive to this, but for now an easy solution is to set the random seed to a value that we know produces feasible test data.

(I initially tried 0 as the seed and all unit test runs failed, then switched to 123 and they all passed)

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Looks good to me. Theory says that a good seed for a RNG is a large, prime number, but I don't think that's super relevant in this case.

Issue #684, opened 2025-03-24
Properly allocate demand to subsectors


tsmbland:
In theory, subsectors are supposed to take demand for a subset of commodities in the sector, and invest only to meet these commodities.

However, this is not what's happening. Subsectors are being passed the full commodity demands for the sector, not just the relevant commodities.

Steps to fix this:
- Modify the `demand_share` module to take a demand object, rather than the full market object
- Filter the market object within the subsector, and pass this filtered object to `self.demand_share` in place of the current `market` object

Issue #389, opened 2024-07-02
Memory error for transport model using scipy solver


tsmbland:
I tried using the scipy solver with @sharwaridixit's transport model, and got this rather alarming error:

`numpy.core._exceptions._ArrayMemoryError: Unable to allocate 40.2 GiB for an array with shape (16, 16, 7, 7, 41, 41, 16, 16) and data type float64`

It's specifically being thrown by [this line](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/4c0b53dc4ca1857689d4a8c177f8b9c691843aad/src/muse/constraints.py#L853) in the calculation of the constraint matrix. I'm guessing there's a bug somewhere that's accidentally duplicating all of the dimensions of an array, but I'd need to investigate further

HarmonicReflux:
Hi @tsmbland,

Can you let me know what command you invoke in the terminal that produces the error, i.e. is it something like 
`muse --model -<your_model_goes_here>`

I take from your description, the culprit is line 853 in src/muse/constraints.py and reads
`reduced = reduce(xr.DataArray.__mul__, diagonal_submats)`

tsmbland:
Hi @HarmonicReflux. To reproduce the error you will need to use the model found [here](https://github.com/EnergySystemsModellingLab/MUSE_OS/issues/338). However, in order to fix the bug described in that issue and get it working with the latest version of MUSE, you will have to replace the settings file with the one attached here:
[settings.zip](https://github.com/user-attachments/files/16174384/settings.zip)
You can then run the model with `muse settings.toml`.

The specific issue I'm describing here only happens when you set `lpsolver` to "scipy" (currently it's using "adhoc").

Let me know if you have any problems. I imagine this one won't be super straightforward to fix.

HarmonicReflux:
- How can one access the source code for the LP solver "adhoc" and where is this solver documented?

- I think the solver "adhoc" acts fundamentally different from scipy's version of solving a linear program, although there
are[ several solvers available ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html) for LP's and settings.toml does not specify which one it uses. In particular there are  "highs", "highs-ds", "interior-point" (legacy), "revised simplex" (legacy), and "simplex" (legacy) available.

In the below, I print a detailed error log:
[...]

PR #687, opened 2025-03-26
Limit search space to relevant technologies

tsmbland:
# Description

Adding some code to `initialize_from_technologies`, which initializes the search space, to only consider technologies that produce at least one demanded commodity. Ultimately, if there are any technologies in the sector that do not produce any demanded commodities, removing them from the search space will shrink the size of the linear problem at no cost.

This will be particularly important once I've done #684 and users can split commodity demands across multiple subsectors. This should then be a viable solution for fixing the memory problems in #389 (watch this space...).

I considered making this a separate filter, but I want it to be included in all models, for all agents, regardless of what filters the user has put in their agents file, so this seemed like the most appropriate place to put it.

Added a test to show that it works.

There was one stubbornly failing test (`test_run_retro_agent`), which I've managed to keep happy in the end. Not too fussed about it though as it's to do with retrofit agents which are on their way out anyway


dalonsoa (approved):
I think it makes perfect sense the way you implemented it. It is surprising this was not already in place - why investing in a technology that does not produce any demanded commodity?

PR #690, opened 2025-03-26
Add tutorial for splitting sector into subsectors

tsmbland:
Added a quick tutorial to demonstrate how a sector can be split into subsectors, and the benefits of doing so.

See the tutorial [here](https://muse-os.readthedocs.io/en/subsectors_tutorial/user-guide/subsectors.html)

~This also nicely demonstrates some of the things that are currently going wrong:~
- ~the first figure should be identical/similar to the figure in [this tutorial](https://muse-os.readthedocs.io/en/subsectors_tutorial/user-guide/additional-service-demand.html) (i.e. without splitting into subsectors), but investments in the residential sector are approximately twice as high~
- ~in the second figure, investment decisions should only be affected for cooking technologies, but heating technologies are also affected~

~I have a fix for all this, but it's messy and not ready to merge yet.~

Edit: rebased off #685 (work in progress fix) and the results look much more sensible now


martinstringer (approved):
This should eventually contain an explanation of how "Quantity" can change for the two service demands once split into subsectors. This is an important feature that it can otherwise take new users some time to realise on their own. But it probably makes sense to merge this version for now and add that additional enhancement later.

I would also vote that at some point, in all the examples, we replace the filenames like ExistingCapacityHeat.csv with the equivalent in lowercase with underscore, like  existing_capacity_heat.csv. This is for exact correspondence with the settings file convention which helps new users keep up with what's going on and is useful if automating input file creation. But I appreciate these reasons are not strong enough to justify making that effort right now.

dalonsoa (approved):
Really nice and super clear tutorial.

I'd add in the final part, after saying that the electric stove is ditched, that the demand for gas has increased, not surprisingly, which might have an impact into the carbon production. Not relevant for the purpose of the tutorial itself, but I think it is good to highlight how the different things and decisions are interrelated.

tsmbland:
> I would also vote that at some point, in all the examples, we replace the filenames like ExistingCapacityHeat.csv with the equivalent in lowercase with underscore, like existing_capacity_heat.csv. This is for exact correspondence with the settings file convention which helps new users keep up with what's going on and is useful if automating input file creation. But I appreciate these reasons are not strong enough to justify making that effort right now.

Yeah unfortunately MUSE is riddled with these types of inconsistencies. We've been much more careful about this in MUSE2

PR #715, opened 2025-06-02
Remove Fuel and EndUse columns from technodata files

### LINKED ISSUE #493: `Fuel` and `EndUse` fields in `Technodata.csv` aren't/shouldn't be required
The purpose of these fields (in theory) is so that the `fueltype` and `same_enduse` search spaces can filter technologies that have the same fuel/enduse.

However, we already know what the fuel and enduse of all the technologies are based on the `CommIn` and `CommOut` files, so these columns in the `Technodata` file shouldn't be necessary.

In fact, `same_enduse` doesn't actually use the `EndUse` field (despite what the documentation says), and determines end-use based on outputs defined in the CommOut file (along with the `is_enduse` filter which excludes environmental commodities)

I suggest we do a similar thing for `fueltype`, and then we can drop both of these fields from the `Technodata` file

tsmbland:
# Description

As explained in the issue, these columns are entirely redundant as this information can be determined from the CommIn and CommOut files. There were a couple of places in the code still using the information from these columns (namely the `same_fuels` filter and the LCOE/EAC outputs), so I've modified these to get the relevant data from the commodity flows instead (via the `comm_usage` field of the technologies). I've also modified the example models and documentation, and added a warning if technodata files still contain these columns (it won't break anything, but it's a good opportunity for users to simplify their input files)

Fixes #493 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -15,134 +15,98 @@ def assets():
     result = Dataset()
     result["year"] = "year", range(2010, 2031)
     assets = set((randint(2010, 2020), choice(list("technology"))) for i in range(30))
-    assets = list(assets)
     result["installed"] = "asset", [u[0] for u in assets]
     result["technology"] = "asset", [u[1] for u in assets]
     shape = len(result.year), len(result.asset)
     result["capacity"] = ("year", "asset"), ones(shape, dtype="int")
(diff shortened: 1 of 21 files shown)


dalonsoa (approved):
The changes look OK to me. The code looks marginally more complex, but it is the right move as it also avoids inconsistencies in the input files.

PR #717, opened 2025-06-02
Tidy scipy adapter code

tsmbland:
Tidying up the scipy adapter code, based on this suggestion:

> I always thought this was a very odd design pattern. Why not simply storing the templates in attributes of ScipyAdapter and making the relevant methods like `back_to_muse` not static? It will be way simpler to follow. 
> All in all, and considering the size of this file, I think it will make sense to extract this as in its own module and make most of these static methods just plain functions in there. They way they are used here feels like the wrong way of using static methods, specially when you are calling them within the class itself.
> _Originally posted by @dalonsoa in https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/685#discussion_r2022191229_

The scipy adapter code now lives in a new file (`lp_adapter.py`), separate from the constraints code, and uses a simpler design pattern, as suggested.

I've also reworked the tests a bit, mostly to tidy up the fixtures and put the tests in a more logical order. I _think_ this is an improvement. Despite splitting the lp adapter tests into a new file, I _haven't_ split the tests into a new file. The reason is that the constraints and lp adapter tests share pretty much the same fixtures, so to avoid duplicating fixture definitions it made more sense to keep the tests in the same file. Maybe there's a better solution though - happy for suggestions


dalonsoa (approved):
I have not gone into the detail of the code, but this is way, way more readable and, specially, the code structure makes sense, facilitating understanding the workflow of the application. I would not worry too much about the structure of the tests.

PR #720, opened 2025-06-03
Tidy `test_readers`

tsmbland:
# Description

I've been having a go with the Cursor AI IDE, and wanted to try it out on the relatively self-contained and harmless task of tidying up some of the tests. The main reasons for this are, firstly, to make it clearer what's actually being tested, and secondly, to make it easer to modify the tests later on if/when I make any changes to the main code. Potentially very useful, but it's such a boring job that I'd never want to do this on my own.

This was my prompt:

> Tidy this test module. Don't reorder any functions or move to other files. I want the code to be more concise, less repetitive, less redundant and more readable. Keep running to make sure the tests all pass

It then spent a few minutes making changes, running the tests, inspecting the error messages, and fixing mistakes until all the tests passed. I had to give it an extra nudge as it initially ignored a chunk of tests in the middle of the file, and some more nudges to keep it running. I then went through the code, made a couple of very minor changes, and asked it to write a summary for me:

### 1. Added Constants for Common Test Values
- Introduced `YEARS`, `COMMODITIES`, and `TIMESLICES` constants to reduce duplication
- Standardized commodity order across tests
- Made commodity order checks order-independent using `set` comparisons

### 2. Improved Test Structure and Organization
- Added clear section comments for setup, verification, and cleanup
- Grouped related assertions logically
- Extracted repeated test data into variables
- Made test data creation more explicit and maintainable

### 3. Enhanced Code Readability
- Added descriptive docstrings to all test functions
- Used consistent formatting and naming conventions
- Improved variable names for clarity
- Simplified complex assertions
- Moved pandas import to top of file with other imports

### 4. Consolidated Related Tests
- Combined similar test cases into single test functions
- Reduced code duplication in utilization factor tests
- Maintained test coverage while reducing code complexity

### 5. Made Tests More Robust
- Made order-dependent checks order-independent
- Improved error messages through better test organization
- Added explicit coordinate consistency checks
- Used consistent patterns for coordinate verification

## Impact
- All tests continue to pass
- No functional changes to the test coverage
- Improved maintainability and readability
- Reduced likelihood of tests breaking due to non-functional changes
- Better organization makes it easier to add new tests in the future

---

Not a particularly difficult task, but I'm actually pretty gobsmacked at how well it's done, with very little manual intervention

One potential pitfall with this approach is that the easiest way for the AI to get all the tests to pass would be for it to simply remove checks. I don't see any evidence of this here, but in general we need to be careful to make sure that everything that was tested before is still being tested. 

Fixes # (issue)

[...]

PR #721, opened 2025-06-04
Tidy tests

tsmbland:
# Description

Following on from #720, I've gone through and used a similar approach to tidy most of the other tests

I just went through each test file one by one and got it to look at them in isolation. I haven't done anything to conftest - this is a much harder problem as the fixtures in here are used by multiple test modules. Something for another time, perhaps.

Definitely not expecting a line-by-line review. I've given it all a glance to make sure there are no glaring issues like test deletions, but it's not practical even for me to scrutinise every change. We can be reassured that all the tests are still passing, and the worst that can happen is that some things that were being tested before are no longer being tested. I'd say it's worth the risk for the benefits of readability and ease of maintenance.

Overall, I think it's done a pretty good job

Observations:
- I found that it occasionally deleted tests. When asking why, it would often give a good reason (e.g. incorporated into another test), but sometimes it would realise it's made a mistake and restore the test
- Sometimes, particularly in large files, it would ignore large chunks of code in the middle of the file and require re-prompting
- The instruction to not reorder functions was to make the diffs easier to inspect. It would usually obey this, but sometimes didn't, in which case I usually had to re-order things manually
- It would sometimes prioritise conciseness over readability, so I had to re-prompt a few times in cases where I felt the code was unreadable or overengineered
- Usually the added inline comments were helpful, but sometimes it would say generic things like "Test edge cases", which needed prompting for clarification. Sometimes it actually removed helpful comments in favour of generic comments like this, which was annoying, so I had to look through carefully and reject the changes whenever it did this
- I rejected the changes to `test_carbon_budget` and `test_objectives` as I didn't like the suggestions, and these are already pretty tidy
- As it inspects error messages, it's useful to make sure error messages within MUSE are as informative as possible. Unfortunately it can't enter debug mode and inspect variables (although it's only a matter of time)

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_agents.py
+++ b/tests/test_agents.py
@@ -15,134 +15,98 @@ def assets():
     result = Dataset()
     result["year"] = "year", range(2010, 2031)
     assets = set((randint(2010, 2020), choice(list("technology"))) for i in range(30))
-    assets = list(assets)
     result["installed"] = "asset", [u[0] for u in assets]
     result["technology"] = "asset", [u[1] for u in assets]
     shape = len(result.year), len(result.asset)
     result["capacity"] = ("year", "asset"), ones(shape, dtype="int")
-    result["maxyears"] = "asset", randint(2010, 2030, len(result.asset))
-    result["capa"] = "asset", randint(1, 100, len(result.asset))
-    result["capacity"] = result.capacity.where(result.year <= result.maxyears, 0)
-    result["capacity"] *= result.capa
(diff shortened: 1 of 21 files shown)


dalonsoa (approved):
I have not done a line by line review, obviously, but this looks broadly OK and the result seems indeed simpler and more readable. As you say, the risks are small.

PR #725, opened 2025-06-12
More comprehensive tests for csv readers

tsmbland:
# Description

I'm in the process of doing some AI-driven test-driven development to improve the input layer, but I want to make sure the tests are absolutely solid so I don't accidentally break anything.

We already had some tests for the reader functions in `test_readers`, but these weren't fully comprehensive, and some readers were missing tests. I've created a new test module `test_csv_readers` with more comprehensive tests

For each xarray object outputted by a reader I'm checking
- dimensions and coordinates
- coordinate values
- data types
- data values for one specific coordinate

I've created schema objects for each xarray, which should give the AI something to work from as I go forward. I created `generate_dataset_schema` and `generate_dataarray_schema` functions to help generate these schemas from xarray objects.

I had to create a new example model `default_correlation` as a few of the input files were not covered by the existing example models

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_csv_readers.py
+++ b/tests/test_csv_readers.py
@@ -49,6 +49,19 @@ class DataArraySchema:
     dtype: str
     name: str
 
+    @classmethod
+    def from_da(cls, data: xr.DataArray) -> DataArraySchema:
+        """Generate a DataArraySchema from an existing DataArray."""
+        return cls(
+            dims=set(data.dims),
+            coords={
+                name: CoordinateSchema(dims=coord.dims, dtype=str(coord.dtype))
+                for name, coord in data.coords.items()
+            },
+            dtype=str(data.dtype),
+            name=data.name,
+        )
+
 
 @dataclass
 class DatasetSchema:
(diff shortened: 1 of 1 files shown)


tests/test_csv_readers.py
```diff
@@ -0,0 +1,857 @@
...
+    expected = {
+        "GDPexp": 0.0994,
+        "constant": 1.01039e-05,
+        "GDPscaleLess": 753.1068725,
+        "GDPscaleGreater": 672.9316672,
+    }
+    assert_single_coordinate(data, coord, expected)
+
+
+def generate_dataarray_schema(data: xr.DataArray) -> DataArraySchema:
```
dalonsoa, line 773: I would make this and the next one class methods for the `DataArraySchema` and the `DatasetSchema`, so they can be used like:

```python
schema = DataArraySchema.from_da(data)
schema = DatasetSchema.from_ds(data)
```

tests/test_csv_readers.py
```diff
@@ -0,0 +1,857 @@
...
+            coords={"commodity": CoordinateSchema(dims=("commodity",), dtype="object")},
+            data_vars={
+                "comm_name": "object",
+                "comm_type": "object",
+                "emmission_factor": "float64",
+                "heat_rate": "int64",
+                "unit": "object",
+            },
+        ),
+    )
```
dalonsoa, line 209: After seeing that you have `generate_dataarray_schema` below, and if you consider my suggestion to make them classmethods, I wonder if it wont be simpler to replace this `assert` schema by:

```python
assert DatasetSchema.from_ds(data) == DatasetSchema(...)
```

It does not save you much - if anything - but as dataclasses have built-in equality comparison, we can very well use them. 

Also, from the point of view of readability, I feel that defining the schema and then doing the comparison would be better:

```python
global_commmodities_schema = DatasetSchema(...)
assert DatasetSchema.from_ds(data) == global_commmodities_schema
```

dalonsoa (approved):
This looks really good and thorough. I've made a couple of comments that I think will help with readability, but not a big deal either way.

PR #727, opened 2025-06-16
Tidy `toml` module

### LINKED ISSUE #713: Issue with case-sensitivity in `excluded_commodities`
MUSE converts commodity names in the csv files to lowercase, but doesn't do this for the `excluded_commodities` list defined in the settings file.

This can cause conflicts if commodity names are not written in lowercase (e.g. "SOLAR")

The simplest thing to do is just to convert the excluded_commodities list to lowercase when we read it in

tsmbland:
# Description

A few misc changes, mostly to trim a bit of fat from the `toml` module. There are a few features here that relate to undocumented (presumably legacy) features, which I'm going to get rid of to make my life easier:
- looks like there was an option to split the toml file across multiple files, but this is undocumented, never seen it used, and looks limited in use as each additional toml file can only have a single section. I could see something like this being useful so people could work on sectors independently and then have a "master toml" which references all the other ones to run the combined simulation. However, we'd need something more fleshed-out that what was there before
- a `muse_sectors` option. Not documented, never seen it used, so it goes
- an ability to specify a directory of global input files, rather than individual paths for `projections` and `global_commodities`. I prefer being more explicit, and again, this isn't documented, so it goes. (However, I still allow these fields to be absent, in which case the default paths defined in `default_settings.toml` are used)
- simplified `format_paths` as some of its features no longer apply

Other notes:
- I've created a few extra settings checks out of existing steps that were previously in `read_settings` or various factory methods. I've renamed these checks to "hooks", as this better captures what they're doing (can modify as as well as validate the settings). I've also moved the decorator function out of `decorators.py` and deleted this module as it didn't make any sense to have a module with just one decorator in it. I've also allowed this to take a priority parameter to determine which order the hooks should be run (e.g. we should check that there are sectors _before_ attempting to sort them)
- Added a `standardise_case` check which fixes #713.
- I'm also trying to be consistent with using pathlib for paths rather than passing around paths as strings. Internally, all paths passed around _should_ now be `Path` objects, although I've tried to ensure that any functions that might be accessed through the API (i.e. when running MUSE from a script) can still take strings as paths.
- Not touching `read_technodata` or `read_presets_sector` - will deal with these in another PR 

This will warrant a bump to v1.5 (hopefully along with #724 if I can get it done soon), and I'll make sure to explain this all in the release notes

Fixes #713

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Very nice. I was involved in some of the things you are dropping - for good reasons - and I can avoid wondering why we did it so badly...

Issue #502, opened 2024-10-01
Allow early decommissioning


tsmbland:
Currently, once an asset is installed, it can never be decommissioned until it reaches its technical life. Commodity demands in a given year will first be met by existing assets operating at full capacity, and then investments are decided on to meet any leftover demand. 

Often, allowing assets to run for their full lifespan is the most sensible cost-minimisation strategy. However, in some cases it may actually be cheaper to decommission assets early and install new assets rather than paying high running costs for existing assets. An example would be if there's a substantial increase in the carbon price that makes high-carbon technologies no longer cost-effective to run.

I imagine this would be require substantial changes to the code, but something we could take into consideration if/when reworking the scipy solver (as discussed in #389)

PR #728, opened 2025-06-16
Calculate prices based only on utilised capacity

tsmbland:
# Description

There's an issue with the calculation of commodity prices for assets with unused capacity

From an exchange with Sharwari:
> The reason that hyg_priv prices spike in 2040 is that consumption (and hence supply) of hyg_priv drops off rapidly in 2040, but there still exists a large amount of capacity due to investment earlier in the simulation when consumption was higher. There are costs associated with maintaining this level of capacity (capital costs and fixed costs), and when supply is low compared to capacity the cost per-unit of output can be very high. Ultimately this raises the LCOE of hyg_priv, causing the price to rise.

In reality, you'd probably end up decommissioning the assets you don't use to avoid having to keep paying the maintenance costs. This is something we plan to introduce in MUSE2, but it's not straightforward to do this in MUSE1 (#502).

However, Adam suggested that we can effectively capture this behaviour, at least from the point of view of price calculations, by calculation the fraction of asset capacity that is _actually utilised_ in a given year, and then use this utilised capacity for the price calculations.

Turned out to be an easy fix as we already have the `capacity_to_service_demand` function which calculates the hypothetical capacity that would be required to meet a set of commodity demands. We can plug the _actual_ supply values for the assets into this, and this will give us the minimum capacity that would have been required to meet this supply i.e. the _utilised capacity_ (which is guaranteed to be less than or equal to the _actual capacity_)

Taking timeslices into account, this will give us the utilised capacity in the _most demanding timeslice_. I.e. if a portion of capacity is used in any timeslice, it is considered utilised for that year.

None of the example models, except for the trade model, are affected by this as they don't suffer from this problem

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
The code changes look OK - although probably tests should be added accordingly. 

Conceptually, it feels odd that the maintenance costs for the whole installation are not reflected into the prices if you just use a small fraction of it, but if Adam is ok with that, so am I, obviously.

ahawkes (approved):
Yes this seems as discussed - thanks!

tsmbland:
@martinstringer Could you please try this out with the UK model?

tsmbland:
I've given this a go with Sharwari's transport model and it seems to make quite a big difference to some commodity prices, for the better.

Surprisingly with Alaa's model there's hardly any effect. A few prices change very slightly but no more than ~1%. I guess it's not suffering from the same problem of under-utilized assets.

ahawkes:
Great thanks.

From: Tom Bland ***@***.***>
Date: Thursday, 19 June 2025 at 12:32
To: EnergySystemsModellingLab/MUSE_OS ***@***.***>
Cc: Hawkes, Adam D ***@***.***>, Review requested ***@***.***>
Subject: Re: [EnergySystemsModellingLab/MUSE_OS] Calculate prices based only on utilised capacity (PR #728)
[Image removed by sender.]tsmbland left a comment (EnergySystemsModellingLab/MUSE_OS#728)<https://github.com/EnergySystemsModellingLab/MUSE_OS/pull/728#issuecomment-2987412703>

I've given this a go with Sharwari's transport model and it seems to make quite a big difference to some commodity prices, for the better.

Surprisingly with Alaa's model there's hardly any effect. A few prices change very slightly but no more than ~1%. I guess it's not suffering from the same problem of under-utilized assets.

tsmbland:
@ahawkes Do we want to do something like this for MUSE2? I've found you can sometimes get crazy prices when using the "full"/"full_average" strategies, particularly in the first year if input assets are under-utilised. Less a problem in later years when unused assets can be (partially) decommissioned early

ahawkes:
Yes, fair enough. Can we raise as an issue for MUSE2, but put on hold for now. Ideally such assets would just get decommissioned, but that might not always happen.

tsmbland:
> Yes, fair enough. Can we raise as an issue for MUSE2, but put on hold for now. Ideally such assets would just get decommissioned, but that might not always happen.

Ok, have opened up https://github.com/EnergySystemsModellingLab/MUSE2/issues/1325

Agree that such assets should be decommissioned. I think this is largely a problem with the first year where prices are calculated before we do any commissioning/decommissioning, based on the assets the user provides in the input file. If capacities in the input data are higher than they need to be then prices may be elevated as a result of this. To some extent that's the fault of the user, but it's not necessarily a clear link and could end up confusing people

PR #730, opened 2025-06-18
More reader tests

tsmbland:
# Description

Adding tests for a few more readers that weren't covered in #725 

We have `read_technologies` and `read_technodata`, which are mostly just wrappers over other reader functions, but there's enough processing going on that they deserve their own tests (particularly in the case of the trade model).

There's also some code in the presets factory which combines the presets data into a dataset. I've moved this into a new function and added a test for it, both for the default model and correlation model as these are a bit different

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_csv_readers.py
+++ b/tests/test_csv_readers.py
@@ -922,7 +922,7 @@ def test_read_technodata(model_path):
             "technology": ["gasCCGT", "windturbine"],
             "region": ["R1"],
             "year": [2020, 2025, 2030, 2035, 2040, 2045, 2050],
-            "comm_usage": [10, 9, 8, 6, 9],
+            "comm_usage": [10, 9, 6, 9],
         },
     )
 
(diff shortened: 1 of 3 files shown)


tests/test_csv_readers.py
```diff
@@ -728,3 +728,367 @@ def test_read_regression_parameters(correlation_model_path):
...
+    }
+    expected = {
+        "consumption": 1.0,
+        "costs": 0,
+        "supply": 0,
+    }
+    assert_single_coordinate(data, coord, expected)
+
+
+def test_read_presets_sector__correlation(correlation_model_path):
```
dalonsoa, line 1046: Why this one requires its own version of the test?

dalonsoa (approved):
Just a small question, but looks OK, otherwise.

tests/test_csv_readers.py
tsmbland: The schemas are exactly the same, but the input data is very different, and if you look at the function it treats these two models very differently, hence needing another test. 

Alternatively, I could have put them both in the same test to avoid duplicating the schema - that might be better actually

PR #738, opened 2025-07-01
Fix rounding issue, include zeros rows for costs/prices

tsmbland:
# Description

this change.*

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


src/muse/outputs/mca.py
```diff
@@ -120,59 +90,28 @@ def factory(
     """
     from muse.outputs.sector import _factory
 
-    def reformat_finite_resources(params):
```
dalonsoa, line 123: I don't think this is included in the PR description. Why is this gone?

PR #739, opened 2025-07-01
Remove units rows from input files

tsmbland:
# Description

We're no longer reading in these rows since #724, so it makes sense to remove them from the example models and documentation. Also dropping the units column from the existing capacity files.

I never liked having a units row as it might give users the impression that they can use any units they want and MUSE will do the conversions, which is not true. 

Some units are still useful to include:
- commodity units (e.g. GW, PJ)
- currency units (see #740)

These are not used by the model, but are useful for documentation purposes. All other parameters (e.g. capital costs, commodity flows) must be expressed in terms of these units (e.g. PJ/Year or USD/PJ), and there is no flexibility here. There's no point having a units row for the other parameters (which users may define incorrectly) when the units can (and must) be worked out with respect to the commodity and currency units

I do still allow the units row to be present, and just ignore it, so existing models will still work fine. I also allow several files to have a units column, which (except for the global commodities file) is also dropped

The documentation still needs a bit of work to make it clear which units users should be using for various parameters. I've started that in #740 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/inputs/commodities_io.rst
+++ b/docs/inputs/commodities_io.rst
@@ -45,7 +45,6 @@ heater.
 .. csv-table:: Commodities used as consumables - Input commodities
    :header: ProcessName, RegionName, Time, Level, electricity
 
-   Unit, -, Year, -, GWh/PJ
    resBoilerElectric, region1, 2010, fixed, 300
    resBoilerElectric, region1, 2030, fixed, 290
 
--- a/docs/inputs/existing_capacity.rst
+++ b/docs/inputs/existing_capacity.rst
@@ -12,10 +12,10 @@ follow the structure reported in the table.
 
 
 .. csv-table:: Existing capacity of technologies: the residential boiler example
-   :header: ProcessName, RegionName, Unit, 2010, 2020, 2030, 2040, 2050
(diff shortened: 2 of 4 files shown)


docs/tutorial-code/add-new-technology/1-introduction/power/CommIn.csv
```diff
@@ -1,5 +1,4 @@
 ProcessName,RegionName,Time,Level,electricity,gas,heat,CO2f,wind,solar
-Unit,-,Year,-,PJ/PJ,PJ/PJ,PJ/PJ,kt/PJ,PJ/PJ,PJ/PJ
 gasCCGT,R1,2020,fixed,0,1.67,0,0,0,0
-windturbine,R1,2020,fixed,0,0,0,0,1,0
-solarPV,R1,2020,fixed,0,0,0,0,0,1
+windturbine,R1,2020,fixed,0,0.0,0,0,1,0
```
tsmbland, line 3: There are a few changes like this. This is just because these files are generated by scripts using pandas. With the units row, all columns were read in as strings. Without the units row, columns are read in as either ints or floats depending on the data, and float columns will always be saved with decimal points

PR #752, opened 2025-07-23
Correctly pass timeslice level to capacity_to_service_demand

tsmbland:
# Description

Tiny bug that was breaking the timeslice_level feature 

Unfortunately, since timeslice_level is an optional parameter for most functions (mostly to make the tests easier to write) it's easy to forget to pass it, but when this happens you can end up with timeslice-mismatch errors

Consider making timeslice_level mandatory to prevent this

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/src/muse/objectives.py
+++ b/src/muse/objectives.py
@@ -227,13 +227,16 @@ def efficiency(
 def capacity_to_service_demand(
     technologies: xr.Dataset,
     demand: xr.DataArray,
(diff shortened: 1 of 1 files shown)


tsmbland:
A few similar issues fixed in bf52b40a584157f72cfa73600d2f419ed80eec14

PR #788, opened 2025-09-23
Don't apply the `with_asset_technology` filter by default

tsmbland:
# Description

There was an observation from Martin/Sharwari of agents investing in technologies that should be excluded by the agent's search rule (in this case it was investing in technologies with a capital cost higher than the agent's spend limit).

I believe this is caused by `with_asset_technology` filter, which allows agents to consider all existing(/just decomissioned) asset technologies for further investment, even if those technologies don't meet the criteria of the other agent filters.

```
@register_filter
def with_asset_technology(
    agent: Agent,
    search_space: xr.DataArray,
    **kwargs,
) -> xr.DataArray:
    """Search space *also* contains its asset technology for each asset."""
    return search_space | (search_space.asset == search_space.replacement)
```

This is applied by default _even if the user doesn't specify this in the input files_. This seems a bit nonsense to me, even for retrofit agents. If a technology doesn't meet the criteria of the agent filters, I don't think we should consider it even if it's in the current asset pool (maybe it was viable at the time of investment, but parameters can change over time so it might not be viable any more).

I've decided to remove the logic in the agents factory that automatically applies this filter, but keep the filter present as an option in `filters.py` (so users can still include it by chaining filters e.g. "spend_limit->with_asset_technology")

### DIFF
--- a/src/muse/agents/factories.py
+++ b/src/muse/agents/factories.py
@@ -71,14 +71,10 @@ def create_retrofit_agent(
 
     kwargs = _standardize_investing_inputs(decision=decision, **kwargs)
 
-    search_rules = kwargs.pop("search_rules")
-    if len(search_rules) < 2 or search_rules[-2] != "with_asset_technology":
-        search_rules.insert(-1, "with_asset_technology")
-
     return InvestingAgent(
         assets=xr.Dataset(dict(capacity=assets)),
         region=region,
-        search_rules=filter_factory(search_rules),
+        search_rules=filter_factory(kwargs.pop("search_rules")),
         year=year,
         **kwargs,
     )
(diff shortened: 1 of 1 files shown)


dalonsoa (approved):
LGTM! The name of the filter is not that clear to me, anyway...

Is there a list of default filters that are applied and that the user can check, maybe in the docs? If so, that one should be updated as well.

tsmbland:
> Is there a list of default filters that are applied and that the user can check, maybe in the docs? If so, that one should be updated as well.

This was never documented in the first place, but after this PR the only default filter applied will be the [`compress`](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/3e6d56317a022fe23ce5c5573d2bc5f503a2f051/src/muse/filters.py#L401) filter, which I don't think users need to be aware of as it's just an optimisation

Apart from that, it will just use what the user puts in the agents file (e.g. "all" or "spend_limit->maturity")

PR #835, opened 2026-04-23
Update prices in the base year

tsmbland:
Prices were previously not being updated in the base year, sticking to the values provided in the projections file (or zero if not provided), whereas all subsequent years return endogenously calculated prices (if available). 

However, I don't think there's a particularly good reason why we _shouldn't_ return endogenous prices for the base year, so in the interests of consistency I've changed this.

I've also tried to clarify the documentation to make it clearer how the projections file is used and the consequences for omitting certain price projections.

You can see the implications of this in the results files for the example models. Apart from changing prices in the first year, this doesn't have any impact, because prices are (with certain exceptions) only used to make investment decisions, and there are no investments in the base year. Still nice to see these numbers, and this will also prevent potentially sharp price changes from the first year to the second year.

With this change there's shouldn't be any need to provide price projections for commodities that can be produced in the model, except that if they don't have a projection they'll end up with a price of zero until the first time they're actually produced.


tsmbland:
@martinstringer FYI

